Issue #268, opened 2026-02-25
sub(#247) Phase 5: Integration tests for backend agent loop


mmogr:
## Goal

Add comprehensive integration tests for the backend agentic loop covering the full cycle: LLM streaming → tool execution → result injection → loop termination.

### 1. Create mock infrastructure

```rust
// tests/common/mock_llm.rs
pub struct MockLlmServer {
    responses: Vec<MockLlmResponse>,
}

pub struct MockLlmResponse {
    pub content: Option<String>,
    pub tool_calls: Vec<ToolCall>,
    pub finish_reason: String,
}

// Configurable: first call returns tool calls, second returns final answer
impl MockLlmServer {
    pub fn new() -> Self { ... }
    pub fn push_response(mut self, response: MockLlmResponse) -> Self { ... }
    pub async fn start(self) -> (String, JoinHandle<()>) { ... }  // Returns URL
}

// tests/common/mock_tools.rs
pub struct MockToolExecutor {
    tools: HashMap<String, Box<dyn Fn(serde_json::Value) -> String>>,
}
```

### 2. Integration tests

```rust
#[tokio::test]
async fn test_simple_tool_call_cycle() {
    // LLM returns 1 tool call → tool executes → LLM returns final answer
}

#[tokio::test]
async fn test_parallel_tool_calls() {
    // LLM returns 3 tool calls → all execute in parallel → LLM returns answer
}

#[tokio::test]
async fn test_max_iterations_reached() {
    // LLM always returns tool calls → hits max iterations → emits error event
}

#[tokio::test]
async fn test_tool_timeout() {
    // Tool hangs → timeout fires → error result returned to LLM
}

#[tokio::test]
async fn test_loop_detection() {
    // LLM keeps calling same tool with same args → loop detected → terminates
}

#[tokio::test]
async fn test_context_budget_pruning() {
    // Messages exceed budget → old tool messages pruned → loop continues
}

#[tokio::test]
async fn test_sse_endpoint_streaming() {
    // Full HTTP integration test: POST /api/agent/chat → receive SSE events
}
```

## Files to Create

| File | Action |
|------|--------|
| `tests/common/mock_llm.rs` | **Create** — Mock LLM HTTP server |
| `tests/common/mock_tools.rs` | **Create** — Mock tool executor |
| `tests/integration_agent_loop.rs` | **Create** — Integration tests |

## Acceptance Criteria

- [ ] Mock LLM server supports multiple scripted responses
[...]

Issue #267, opened 2026-02-25
sub(#247) Phase 4: CLI chat --agent interactive mode


mmogr:
## Goal

Add a `gglib chat --agent` CLI mode that runs the backend agentic loop interactively in the terminal, enabling tool-calling conversations without a browser.

## Background

Phase 3 (#266) exposes the agent loop via HTTP. This phase adds a CLI subcommand that uses the same loop directly (without going through HTTP).

### 1. Add `chat --agent` subcommand to gglib-cli

```rust
// crates/gglib-cli/src/commands/chat.rs
#[derive(Parser)]
pub struct ChatArgs {
    /// Enable agentic mode with tool calling
    #[arg(long)]
    agent: bool,
    
    /// Maximum iterations for agent loop
    #[arg(long, default_value = "25")]
    max_iterations: usize,
    
    /// Enable specific tools (comma-separated, or "all")
    #[arg(long, default_value = "all")]
    tools: String,
    
    /// System prompt
    #[arg(long)]
    system: Option<String>,
}
```

### 2. Implement interactive CLI agent

```rust
pub async fn run_agent_chat(args: &ChatArgs, state: &AppState) -> Result<()> {
    let config = AgentConfig {
        max_iterations: args.max_iterations,
        ..Default::default()
    };
    
    let tool_executor = CliToolExecutor::new(state);
    let agent = AgentLoop::new(state.proxy_url.clone(), Arc::new(tool_executor));
    
    let mut messages = vec![];
    if let Some(system) = &args.system {
        messages.push(json!({"role": "system", "content": system}));
    }
    
    loop {
        // Read user input
        let input = read_user_input()?;
        if input == "/quit" { break; }
        
        messages.push(json!({"role": "user", "content": input}));
        
        // Run agent loop with terminal output
        let (tx, mut rx) = tokio::sync::mpsc::channel(100);
        let agent_clone = agent.clone();
        let messages_clone = messages.clone();
        let config_clone = config.clone();
        
        tokio::spawn(async move {
            agent_clone.run(messages_clone, config_clone, tx).await
        });
        
        // Stream events to terminal
        while let Some(event) = rx.recv().await {
            match event {
                AgentEvent::TextDelta { content } => print!("{}", content),
                AgentEvent::ToolCallStart { tool_call } => {
                    println!("\n🔧 Calling: {} ...", tool_call.name);
                }
                AgentEvent::ToolCallComplete { result } => {
                    println!("   ✅ {} ({}ms)", 
[...]

Issue #266, opened 2026-02-25
sub(#247) Phase 3: Axum SSE endpoint for agentic chat


mmogr:
## Goal

Expose the backend agentic loop via an Axum SSE endpoint so the frontend (and other clients) can use server-side agent execution.

## Background

Phase 2 (#265) implements the core agentic loop. This phase exposes it as an HTTP endpoint with SSE streaming for real-time progress.

### 1. Add SSE route in `gglib-axum`

```rust
// POST /api/agent/chat — starts an agentic conversation
// Returns: SSE stream of AgentEvent
async fn agent_chat(
    State(state): State<AppState>,
    Json(body): Json<AgentChatRequest>,
) -> Sse<impl Stream<Item = Result<Event, Infallible>>> {
    let (tx, rx) = tokio::sync::mpsc::channel::<AgentEvent>(100);
    
    tokio::spawn(async move {
        let agent = AgentLoop::new(state.proxy_url, state.tool_executor);
        agent.run(body.messages, body.config.unwrap_or_default(), tx).await
    });
    
    let stream = ReceiverStream::new(rx).map(|event| {
        Ok(Event::default()
            .event(event.event_type())
            .json_data(&event)
            .unwrap())
    });
    
    Sse::new(stream)
}

#[derive(Deserialize)]
struct AgentChatRequest {
    messages: Vec<serde_json::Value>,
    config: Option<AgentConfig>,
    tool_filter: Option<Vec<String>>,  // Optional: only enable specific tools
}
```

### 2. Define SSE event types

```
event: thinking
data: {"content": "Let me search for that..."}

event: text_delta  
data: {"content": "Based on my research, "}

event: tool_start
data: {"tool_call": {"id": "tc_1", "name": "web_search", "arguments": {...}}}

event: tool_complete
data: {"result": {"tool_call_id": "tc_1", "content": "...", "duration_ms": 1200}}

event: iteration
data: {"iteration": 2, "tool_calls": 3}

event: final_answer
data: {"content": "Here is the complete answer..."}

event: error
data: {"message": "Max iterations reached"}
```

### 3. Add tool executor adapter

Create an adapter that bridges the `ToolExecutorPort` trait to the existing tool infrastructure:

```rust
pub struct HttpToolExecutor {
    mcp_manager: Arc<McpManager>,
    builtin_tools: Arc<BuiltinToolRegistry>,
}

#[async_trait]
impl ToolExecutorPort for HttpToolExecutor {
[...]

Issue #265, opened 2026-02-25
sub(#247) Phase 2: Implement the core agentic loop in Rust


mmogr:
## Goal

Implement the core agentic loop in Rust using the domain types from Phase 1. This loop drives the LLM → tool → LLM cycle server-side, mirroring the frontend's ReAct-lite protocol.

## Background

Phase 1 (#264) defines the types and traits. This phase implements the actual loop logic.

### 1. Create `AgentLoop` adapter in `gglib-axum`

```rust
// crates/gglib-axum/src/agent_loop.rs
pub struct AgentLoop {
    proxy_url: String,
    tool_executor: Arc<dyn ToolExecutorPort>,
}

impl AgentLoop {
    pub async fn run(
        &self,
        mut messages: Vec<serde_json::Value>,
        config: AgentConfig,
        tx: tokio::sync::mpsc::Sender<AgentEvent>,
    ) -> Result<String, anyhow::Error> {
        let mut state = AgentLoopState::new(&config);
        
        loop {
            if state.iteration >= config.max_iterations {
                tx.send(AgentEvent::Error { 
                    message: format!("Max iterations ({}) reached", config.max_iterations) 
                }).await?;
                break;
            }
            
            // 1. Call LLM with tools
            let tools = self.tool_executor.list_tools().await;
            let response = self.call_llm(&messages, &tools).await?;
            
            // 2. Check for text content
            if let Some(content) = &response.content {
                tx.send(AgentEvent::TextDelta { content: content.clone() }).await?;
            }
            
            // 3. Check for tool calls
            if response.tool_calls.is_empty() {
                tx.send(AgentEvent::FinalAnswer { 
                    content: response.content.unwrap_or_default() 
                }).await?;
                return Ok(response.content.unwrap_or_default());
            }
            
            // 4. Execute tools in parallel
            let results = self.execute_tools_parallel(
                &response.tool_calls, &config, &tx
            ).await;
            
            // 5. Append tool results to messages
            messages.push(assistant_message_with_tool_calls(&response));
            for result in &results {
                messages.push(tool_result_message(result));
            }
            
            // 6. Loop detection and budget management
            state.iteration += 1;
            self.prune_for_budget(&mut messages, &config);
        }
        
        Err(anyhow::anyhow!("Agent loop terminated without final answer"))
    }
}
```

### 2. Implement parallel tool execution

```rust
async fn execute_tools_parallel(
    &self,
    tool_calls: &[ToolCall],
[...]

PR #308, opened 2026-02-27
feat(agent): Backend-side agentic loop for CLI and headless tool execution (#247)

### LINKED ISSUE #247: feat(agent): Implement backend-side agentic loop for CLI and headless tool execution
## Summary

All tool execution currently happens client-side in the React frontend (`runAgenticLoop.ts`). The Rust backend (gglib-proxy, gglib-axum) passes tool definitions through to the LLM but never executes tools or drives the agentic loop itself.

This means:
- **CLI mode** cannot do agentic tool calling
- **Web-only mode** (Axum without frontend) cannot drive tool loops
- **Headless deployments** require a browser to use tools
- The frontend orchestrates complex multi-turn LLM interactions, which is fragile for long-running tasks

## Current Architecture

```
Frontend (React) ─── HTTP/SSE ──→ gglib-axum ──→ gglib-proxy ──→ llama-server
     ↕                                                              ↕
Tool Registry                                                  Tool responses
     ↕                                                         (streamed back)
MCP Client (TS)
```

The frontend:
1. Sends chat request with tool definitions
2. Receives streaming response
3. Detects `finish_reason: "tool_calls"`
4. Executes tools via `ToolRegistry` / MCP bridge
5. Sends results back as tool messages
6. Repeats until `finish_reason: "stop"`

This entire loop should be available server-side.

## Proposed Architecture

```
Frontend (React) ─── HTTP/SSE ──→ gglib-axum ──→ gglib-agent ──→ gglib-proxy ──→ llama-server
     ↕                                ↕              ↕
  UI updates             SSE progress events    Tool execution
  (passive)                                        ↕
                                              gglib-mcp (Rust)
                                              (already exists)
```

### Phase 1: Core agent loop crate (`gglib-agent`)

New crate with:
- `AgentLoop` struct with configurable max iterations, tool definitions, and tool executor trait
- Loop logic: call LLM → check finish_reason → execute tools → repeat
- Tool executor trait allowing pluggable backends (MCP, built-in, custom)
- Working memory / context pruning (port logic from frontend `agentLoop.ts`)
- Loop detection (port from frontend `checkToolLoop`)

### Phase 2: SSE progress events

Extend the chat API SSE stream with agent-specific events:
```jsonl
data: {"type":"agent.tool_start","tool":"mcp_search","args":{...},"iteration":2}
data: {"type":"agent.tool_result","tool":"mcp_search","success":true,"duration_ms":340}
data: {"type":"agent.iteration","iteration":3,"total_tool_calls":5}
data: {"type":"agent.content","delta":"Based on the search results..."}
```

Frontend can render these as real-time progress indicators.

### Phase 3: Wire into Axum routes

New endpoint or parameter on existing `/api/chat`:
```
POST /api/chat?agent=true
```
or
```
POST /api/agent/chat
```

When agent mode is requested, the backend drives the loop instead of returning tool_calls to the frontend.

### Phase 4: CLI agentic mode

```bash
gglib chat --agent --tools mcp:filesystem,builtin:time "Find all TODO comments in src/"
```

### Phase 5: Integration tests

Test the full loop with mock LLM responses and mock tool executors.

## Key Design Decisions

1. **Reuse `gglib-mcp` for tool execution** — the Rust MCP client already supports `initialize`, `list_tools`, `call_tool` over stdio JSON-RPC
2. **Frontend becomes passive viewer** — when backend agent mode is active, the frontend just renders SSE events rather than orchestrating
3. **Backward compatible** — default behavior (no `agent` param) still sends tool_calls to the frontend as before
4. **Tool executor trait** for extensibility:
```rust
#[async_trait]
pub trait ToolExecutor: Send + Sync {
    async fn execute(&self, name: &str, args: serde_json::Value) -> ToolResult;
    fn available_tools(&self) -> Vec<ToolDefinition>;
}
```

## Files to Create/Modify

| File | Change |
|------|--------|
| `crates/gglib-agent/` | New crate: agent loop, tool executor trait, working memory |
| `crates/gglib-axum/src/chat_api.rs` | Add agent mode support to chat endpoint |
| `crates/gglib-axum/src/routes.rs` | Optional: new `/api/agent/chat` route |
| `crates/gglib-cli/src/` | Add `--agent` flag to chat command |
| `crates/gglib-mcp/src/` | Implement `ToolExecutor` trait for MCP client |
| `Cargo.toml` | Add `gglib-agent` to workspace |

## Acceptance Criteria

- [ ] Backend can autonomously drive multi-turn tool-calling loops
- [ ] SSE progress events stream tool execution status to connected clients
- [ ] CLI can run agentic conversations with tool support
- [ ] MCP tools work identically whether executed frontend-side or backend-side
- [ ] Loop detection, stagnation detection, and max iteration limits enforced
- [ ] Context pruning prevents context window overflow in long agent sessions
- [ ] Existing frontend-driven tool calling continues to work (backward compatible)
- [ ] Integration tests with mock LLM and mock tools

mmogr:
## Summary

Implements all five phases of #247: a complete server-side agentic loop that brings multi-turn tool-calling to CLI and headless deployments without requiring a browser.

**Before:** the React frontend owned the entire LLM→tool→LLM cycle. CLI users couldn't use tools; web-only mode had no agentic capability; long-running tasks required a persistent browser tab.

**After:** the backend drives the loop. The frontend (when present) is a passive SSE consumer. CLI users get an interactive `--agent` REPL. The same loop logic runs everywhere.

---

### Phase 1 — Core domain types (`gglib-core`)
- `domain/agent.rs` — `AgentConfig`, `ToolCall`, `ToolResult`, `AgentEvent`, `AgentMessage`, `LlmStreamEvent`, `ToolDefinition`
- `ports/agent.rs` — `AgentLoopPort`, `ToolExecutorPort` traits (hexagonal architecture)
- `ports/llm_completion.rs` — `LlmCompletionPort` trait

### Phase 2 — Core agentic loop (`gglib-agent`, new crate)
- `agent_loop.rs` — `AgentLoop` implements `AgentLoopPort`; stateless, re-entrant
- `stream_collector.rs` — consumes `LlmStreamEvent` stream, forwards `TextDelta` live, assembles `ToolCall`s from deltas
- `tool_execution.rs` — parallel dispatch with `Semaphore` + per-tool `tokio::time::timeout`
- `loop_detection.rs` — FNV-1a batch signatures, direct port of `agentLoop.ts` algorithm
- `stagnation.rs` — repeated-text guard using the same FNV-1a hash
- `context_pruning.rs` — two-pass budget trim (tool-message prune → tail prune), port of `pruneForContextBudget`
- `filter.rs` — `FilteredToolExecutor` allowlist wrapper

### Phase 3 — Axum SSE endpoint (`gglib-axum`)
- `POST /api/agent/chat` — streams `AgentEvent` as SSE; `AbortOnDrop` RAII guard cancels the background task when the client disconnects
- `LlmCompletionAdapter` (`gglib-runtime`) — bridges `LlmCompletionPort` to the llama-server SSE stream
- `McpToolExecutorAdapter` (`gglib-mcp`) — bridges `ToolExecutorPort` to the existing MCP client

### Phase 4 — CLI REPL (`gglib-cli`)
- `gglib chat --agent` — interactive multi-turn agent conversation in the terminal
- `rustyline` for readline with history; blocking I/O runs in `spawn_blocking`
- `tokio::select!` on `ctrl_c()` cancels the current turn without exiting
- `--max-iterations`, `--tool-filter`, `--system-prompt` flags

### Phase 5 — Integration tests (`gglib-agent/tests/`)
- `MockLlmPort` — scripted response queue; no HTTP server needed
- `MockToolExecutorPort` — per-tool `Immediate` / `Delayed` / `Failing` / `Counter` behaviours
- Six tests covering: simple cycle, parallel tools, max iterations, tool timeout, loop detection, context budget pruning

---

## Architecture change

```
Before:
  Frontend (React) ─── HTTP/SSE ──→ gglib-axum ──→ gglib-proxy ──→ llama-server
       ↕
  TS Tool registry / MCP client

After:
  Frontend (React) ─── HTTP/SSE ──→ gglib-axum ──→ gglib-agent ──→ gglib-runtime ──→ llama-server
  (passive viewer)                        ↓               ↓
                              SSE progress events    gglib-mcp (Rust tool executor)
```

---

## Acceptance criteria from #247

- [x] Backend autonomously drives multi-turn tool-calling loops
- [x] SSE progress events stream tool execution status (`tool_call_start`, `tool_call_complete`, `iteration_complete`, `final_answer`, `error`)
- [x] CLI can run agentic conversations with tool support (`gglib chat --agent`)
- [x] MCP tools work identically whether executed frontend-side or backend-side
- [x] Loop detection, stagnation detection, and max iteration limits enforced
- [x] Context pruning prevents context window overflow in long agent sessions
- [x] Existing frontend-driven tool calling continues to work (new `/api/agent/chat` endpoint; existing `/api/chat` untouched)
- [x] Integration tests with mock LLM and mock tools (all in-process, ~seconds)

---

## Closes

Closes #247
Sub-issues: #264, #265, #266, #267, #268


crates/gglib-agent/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+path = "src/lib.rs"
+
+[dependencies]
+# Domain types and port traits — the ONLY workspace crate this crate depends on.
+# Hexagonal boundary: gglib-agent must not reference any infrastructure crate.
+gglib-core   = { path = "../gglib-core" }
+tokio        = { workspace = true }
+futures-util = "0.3"
+futures-core = "0.3"
+async-trait  = { workspace = true }
```
mmogr: Both `futures-util` and `futures-core` use bare version strings here instead of `workspace = true`. Every other dependency in this file uses workspace inheritance. This risks version divergence if the workspace root is updated.

Add to `[workspace.dependencies]` in the root `Cargo.toml`:
```toml
futures-util = { version = "0.3", default-features = false }
futures-core = { version = "0.3", default-features = false }
```
Then change these lines to:
```toml
futures-util = { workspace = true }
futures-core = { workspace = true }
```

crates/gglib-agent/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+gglib-core    = { path = "../gglib-core" }
+tokio         = { workspace = true, features = ["macros", "rt-multi-thread"] }
+serde_json    = { workspace = true }
+futures-util  = "0.3"
+futures-core  = "0.3"
+async-trait   = { workspace = true }
+anyhow        = { workspace = true }
+
+[lints]
+workspace = true
```
mmogr: These `[dev-dependencies]` entries are redundant. `gglib-core`, `serde_json`, `futures-util`, `futures-core`, `async-trait`, and `anyhow` are already in `[dependencies]`, which makes them automatically available to `#[cfg(test)]` modules within the same crate. The entire `[dev-dependencies]` block can be reduced to just:
```toml
[dev-dependencies]
tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
```
All other dev-deps are the same as production-deps and will be resolved from there.

crates/gglib-agent/src/context_pruning.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+        AgentMessage::Tool {
+            tool_call_id,
+            content,
+        } => tool_call_id.len() + content.len(),
+    }
+}
+
+/// Total estimated character count across all messages.
+pub fn total_chars(messages: &[AgentMessage]) -> usize {
```
mmogr: `total_chars` is only used within this crate (by `prune_for_budget` and inline tests). Exporting it as `pub` leaks an implementation detail as public API.

Change to `pub(crate)`:
```rust
pub(crate) fn total_chars(messages: &[AgentMessage]) -> usize {
```

crates/gglib-agent/src/agent_loop.rs
```diff
@@ -0,0 +1,477 @@
...
+                    .await;
+
+            // ---- 8. Append assistant + tool-result messages -----------------
+            messages.push(AgentMessage::Assistant {
+                content: if response.content.is_empty() {
+                    None
+                } else {
+                    Some(response.content)
+                },
+                tool_calls: Some(response.tool_calls.clone()),
```
mmogr: This `.clone()` is unnecessary. `execute_tools_parallel` in step 7 only borrows `&response.tool_calls`; that borrow is fully released by the time we reach step 8. The `response` struct can be partially moved here:

```rust
messages.push(AgentMessage::Assistant {
    content: if response.content.is_empty() {
        None
    } else {
        Some(response.content)  // moves content
    },
    tool_calls: Some(response.tool_calls),  // moves tool_calls — no clone needed
});
```

The compiler will ensure both fields are moved only once. This avoids a heap allocation (a `Vec<ToolCall>` copy) on every iteration of the loop.

crates/gglib-agent/src/loop_detection.rs
```diff
@@ -0,0 +1,300 @@
...
+            *n += 1;
+            *n
+        } else {
+            self.hits.insert(sig.clone(), 1);
+            1
+        };
+        if count > max_strikes {
+            return Err(AgentError::LoopDetected { signature: sig });
+        }
+        Ok(())
```
mmogr: The comment correctly notes the optimisation, but the reasoning doesn't hold. `batch_signature(calls)` already allocates a new `String` on every call to `check()`, so `sig` is freshly heap-allocated unconditionally — the entry-API clone of an already-owned value is the same cost as the `.clone()` in the else branch here.

The idiomatic equivalent:
```rust
let count = *self.hits.entry(sig).and_modify(|n| *n += 1).or_insert(1);
```
is six lines shorter, doesn't require the comment explaining the non-idiomatic pattern, and reads as "insert 1 or increment" at a glance. Consider simplifying.

crates/gglib-agent/src/stagnation.rs
```diff
@@ -0,0 +1,159 @@
...
+                        count = self.count,
+                    )));
+                }
+            }
+            _ => {
+                self.count = 0;
+                self.prev_hash = Some(hash);
+            }
+        }
+        Ok(())
```
mmogr: The semantics here are subtly off from the parameter name. The **first** occurrence sets `prev_hash` and leaves `count = 0`. The counter is only incremented on a *repetition*. With `max_stagnation_steps = 5` (the default), stagnation fires on the **6th** consecutive identical response — after 5 repeats of the first occurrence.

Three options, pick the clearest:

**A** — Rename field to `max_stagnation_repeats` and update doc ("abort after this many *repeats* of the same text"):
```rust
/// Abort after this many consecutive *repeats* of the same response text.
/// The first occurrence is always free; the (N+1)th identical response triggers the error.
pub max_stagnation_repeats: usize,
```

**B** — Start `count = 1` on the first occurrence:
```rust
_ => {
    self.count = 1;   // first occurrence counts as 1
    self.prev_hash = Some(hash);
}
```
With this, `max_stagnation_steps = 5` means "abort if the same text has been seen 5 times total" — which matches the name.

**C** — Change the condition to `> max_steps` so "steps" means the number of *repeats* allowed before aborting (current behaviour with clearer naming).

Whatever is chosen, the `AgentConfig` field docs and `StagnationDetector::record` docs should explicitly state "N means __ consecutive identical __ before aborting".

src/hooks/useGglibRuntime/streamAgentChat.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  // ── Create the first assistant message ───────────────────────────────────
+  let iteration = 1;
+  const makeNextMessage = (): string => {
+    const msg = mkAssistantMessage({ turnId, iteration, conversationId });
+    setMessages(prev => [...prev, msg]);
+    setCurrentStreamingAssistantMessageId?.(msg.id!);
+    return msg.id!;
+  };
+
```
mmogr: This closure captures the outer mutable `iteration` variable implicitly. The data flow is easy to miss: `iteration` is mutated in the `iteration_complete` case and then read silently here on the next call. 

Consider making the dependency explicit by taking `iteration` as a parameter:

```ts
const makeNextMessage = (iter: number): string => {
  const msg = mkAssistantMessage({ turnId, iteration: iter, conversationId });
  setMessages(prev => [...prev, msg]);
  setCurrentStreamingAssistantMessageId?.(msg.id!);
  return msg.id!;
};

// Initial call:
let currentId = makeNextMessage(1);

// In iteration_complete:
currentId = makeNextMessage(event.iteration + 1);
```

This also eliminates the `let iteration = 1` mutable variable entirely — `iteration` becomes a transient local in each call site, which is cleaner.

tests/ts/hooks/useGglibRuntime/wireMessages.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    ]);
+    expect(wire).toEqual<AgentWireMessage[]>([
+      { role: 'user', content: 'Hello world' },
+    ]);
+  });
+
+  it('strips non-text parts when building user content string', () => {
+    const wire = convertToWireMessages([
+      userMsg([
+        { type: 'image', image: 'data:image/png;base64,abc' } as never,
```
mmogr: `as never` is a type-system escape hatch that bypasses all checking — it's stronger than `as unknown` and is typically a sign of a type modelling gap. Here the test is asserting that `convertToWireMessages` ignores non-text parts of content arrays (image parts, etc.), which is a legitimate test scenario.

Better approach: add an image type to `GglibMessagePart` if image parts genuinely appear in the UI's message store, or cast as `unknown as GglibMessagePart[]`:

```ts
userMsg([
  { type: 'image', image: 'data:image/png;base64,abc' } as unknown as GglibMessagePart,
  { type: 'text', text: 'describe this' },
] as GglibMessagePart[]),
```

`as never` is especially dangerous here because TypeScript won't catch if the test fixture becomes invalid when types change.

crates/gglib-axum/src/handlers/agent.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// etc.).
+    pub config: Option<AgentConfig>,
+
+    /// Optional allowlist of tool names to expose to the model.
+    ///
+    /// When `Some`, only tools whose names appear in this list are sent to the
+    /// LLM and can be executed. When `None`, all tools from all connected MCP
+    /// servers are available.
+    pub tool_filter: Option<Vec<String>>,
+}
```
mmogr: The distinction between `tool_filter: None` ("all tools allowed") and `tool_filter: Some([])` ("no tools allowed") is only documented in the TypeScript layer. The Rust doc comment here says "When `None`, all tools from all connected MCP servers are available" which is correct, but doesn't make the `Some([])` → "no tools" case equally explicit.

Consider:
```rust
/// - `None`: all tools from connected MCP servers are available
/// - `Some([])`: no tools are exposed (effectively disables tool calling)
/// - `Some(["tool_a", "tool_b"])`: only the listed tools are exposed
```

This matters because `null` and `[]` serialise differently in JSON (a client passing an empty array intending "default all tools" would silently get "no tools").

crates/gglib-agent/src/stream_collector.rs
```diff
@@ -0,0 +1,324 @@
...
+                });
+            }
+        }
+    }
+
+    // The stream ended without a Done event — treat as an infrastructure error.
+    anyhow::bail!("LLM stream ended without a Done event")
+}
+
+// =============================================================================
```
mmogr, line 222: When the stream yields no items at all (immediately empty), this fires the same error message as when the stream terminates mid-response. Both are reported as `AgentError::Internal("stream collection error: LLM stream ended without a Done event")` to the caller. Consider enriching the message or making it a separate error variant so you can distinguish "stream never started" (connectivity issue?) from "stream terminated early" (truncated response) in logs/metrics.

src/hooks/useGglibRuntime/wireMessages.ts
```diff
@@ -0,0 +1,95 @@
...
+      const content = Array.isArray(msg.content)
+        ? (msg.content as { type: string; text?: string }[])
+            .filter(p => p.type === 'text')
+            .map(p => p.text ?? '')
+            .join('')
+        : (msg.content as string) ?? '';
+      result.push({ role: msg.role, content });
+
+    } else if (msg.role === 'assistant') {
+      const parts = Array.isArray(msg.content)
```
mmogr: The inline type assertion here:
```ts
(msg.content as { type: string; toolCallId?: string; toolName?: string; args?: unknown; argsText?: string; result?: unknown }[])
```
is a red flag that `GglibMessage['content']` isn't typed as `GglibMessagePart[]` — if it were, accessing `.toolCallId`, `.toolName`, etc. on `ToolCallPart` would work without casting.

This PR introduces `GglibToolCallPart` and `GglibMessagePart` in `messages.ts` — these new types should be applied here so the conversion function works with proper types instead of anonymous object shapes. The `parts.filter(p => p.type === 'tool-call')` should return `GglibToolCallPart[]` which exposes all the fields directly.

Concretely: change the message type so `GglibMessage['content']` for `role === 'assistant'` is typed as `GglibMessagePart[]` and use the new types here instead of the inline assertion.

mmogr (commented):
Review complete. See inline comments for specifics. Key items to address before merge: workspace version consistency for `futures-util`/`futures-core`, redundant dev-dependencies, `total_chars` visibility, and the avoidable `.clone()`. The stagnation semantics and `makeNextMessage` closure coupling are worth cleaning up for long-term clarity.

crates/gglib-runtime/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde_json = { workspace = true }
 chrono = { workspace = true }
 
 # Async utilities
+futures-core = "0.3"
 futures-util = "0.3"
```
mmogr: **[Medium — DRY violation: workspace dep pins not used]**

These two lines use bare version strings instead of `workspace = true`:

```toml
futures-core = "0.3"
futures-util = "0.3"
```

The workspace `Cargo.toml` explicitly added `futures-util` and `futures-core` as workspace deps in this very PR (with the comment *"Both crates are published in lock-step from the same futures-rs release; using the same constraint here ensures Cargo always resolves them to the same version"*) — but `gglib-runtime` skipped the memo.

Fix:
```toml
futures-core = { workspace = true }
futures-util = { workspace = true }
```

Same pattern applies to `reqwest = { version = "0.13", … }` on line 27 — there is already a `reqwest = "0.13.2"` workspace dep; use it here.

src-tauri/src/main.rs
```diff
@@ -114,6 +114,7 @@ fn main() {
                 hf_client: ctx.hf_client.clone(),
                 runner: ctx.runner.clone(),
```
mmogr, line 115: **[Medium — two separate `reqwest::Client::new()` calls across app boot]**

`src-tauri/src/main.rs` now calls `reqwest::Client::new()` at line ~115 and stores it in `AxumContext::http_client`. The `LlmCompletionAdapter` in the Axum handler then clones it correctly.

However, `gglib-cli` (and possibly the Tauri binary's non-Axum path) each allocate a brand-new `reqwest::Client` inside `LlmCompletionAdapter::new()` for every agent request. Every `Client::new()` creates an entirely fresh connection pool.

For the CLI interactive REPL this is one pool per agent turn, which is unnecessary overhead and defeats HTTP keep-alive across turns. `LlmCompletionAdapter` already supports `with_client(port, client, model)` — the CLI entrypoint should construct one `Client` at startup and share it across turns.

This is worth fixing before the CLI REPL is used more heavily.

crates/gglib-agent/src/loop_detection.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// before erroring on the third, matching the frontend's
+    /// `MAX_SAME_SIGNATURE_HITS = 2` behaviour.
+    pub(crate) fn check(&mut self, calls: &[ToolCall], max_strikes: usize) -> Result<(), AgentError> {
+        let sig = batch_signature(calls);
+        let count = *self.hits.entry(sig.clone()).and_modify(|n| *n += 1).or_insert(1);
+        if count > max_strikes {
+            return Err(AgentError::LoopDetected { signature: sig });
+        }
+        Ok(())
+    }
```
mmogr: **[Low — non-idiomatic map entry pattern]**

```rust
let count = *self
    .hits
    .entry(sig.clone())
    .and_modify(|n| *n += 1)
    .or_insert(1);
```

The `and_modify` + `or_insert(1)` idiom is subtle — the modifier increments existing entries, and new entries start at 1 — but a first-time reader has to trace both branches mentally to confirm no off-by-one exists.

The three-liner is equivalent and more readable:

```rust
let entry = self.hits.entry(sig.clone()).or_insert(0);
*entry += 1;
let count = *entry;
```

crates/gglib-agent/src/context_pruning.rs
```diff
@@ -0,0 +1,524 @@
...
+            content: s.to_owned(),
+        }
+    }
+    fn user(s: &str) -> AgentMessage {
+        AgentMessage::User {
+            content: s.to_owned(),
+        }
+    }
+    fn assistant_text(s: &str) -> AgentMessage {
+        AgentMessage::Assistant {
```
mmogr: **[Low — double blank line before `#[cfg(test)]`]**

Two consecutive blank lines here. `rustfmt` enforces a maximum of one blank line between items. This will cause `cargo fmt --check` to report a diff. Remove one blank line.

crates/gglib-agent/src/tool_execution/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+//!
+//! - All tool calls in a batch are dispatched concurrently via a
+//!   [`tokio::task::JoinSet`].  When the future returned by
+//!   [`execute_tools_parallel`] is dropped (e.g. because `AgentTaskGuard`
+//!   aborts the parent agent task on client disconnect), the `JoinSet` is
+//!   dropped and every in-flight sub-task is cancelled — no resource leak.
+//! - A [`tokio::sync::Semaphore`] caps the number of *simultaneously running*
+//!   tool calls at [`AgentConfig::max_parallel_tools`].
+//! - Each call is wrapped in a [`tokio::time::timeout`] capped at
+//!   [`AgentConfig::tool_timeout_ms`].
```
mmogr, line 13: **[Low — semaphore concurrency cap is not tested]**

The `Semaphore(config.max_parallel_tools)` cap is the core correctness guarantee of this function, but there is no test that verifies tools beyond the cap are actually queued until a slot frees up.

The current integration tests exercise parallel happy-path and single-tool timeout, but both use tools with immediate returns, so the semaphore's `acquire_owned().await` never actually blocks.

A test along these lines would complete the picture:

```rust
// Given max_parallel_tools = 1 and two tools that each take 50 ms,
// their total wall-clock time should be >= 100 ms (sequential),
// not ~50 ms (parallel).
```

Consider adding this to `tests/integration_agent_execution.rs` or a new `tests/unit_tool_execution.rs`.

crates/gglib-agent/src/stream_collector.rs
```diff
@@ -0,0 +1,460 @@
+//! Streaming LLM response collector.
```
mmogr, line 1: **[Low — large inline test block; DoS-guard path not tested]**

Two separate observations here:

1. **Missing test for `MAX_TOOL_CALL_INDEX` guard.** The guard at `index > 64` that prevents DoS via a stream with thousands of tool-call slots is documented in the module, but there is no test that passes a stream with `index = 65` and asserts the appropriate error. This would complete the DoS-protection story.

2. **Consider extracting the test block.** At 507 lines of which ~220 are tests, this is the largest file in the crate. Moving `#[cfg(test)] mod tests { … }` to `tests/unit_stream_collector.rs` would align with the pattern used in `agent_loop.rs` and allow the tests to share the richer `tests/common/` mock infrastructure.

.github/workflows/badges.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           fi
           
           # --- Per-crate Rust tests ---
-          for CRATE in gglib-core gglib-db gglib-gguf gglib-hf gglib-download gglib-mcp gglib-proxy gglib-runtime gglib-gui gglib-cli gglib-axum gglib-tauri gglib-build-info gglib-voice; do
+          for CRATE in gglib-core gglib-db gglib-gguf gglib-hf gglib-download gglib-mcp gglib-agent gglib-proxy gglib-runtime gglib-gui gglib-cli gglib-axum gglib-tauri gglib-build-info gglib-voice; do
```
mmogr: **[Low — crate list duplicated 6 times in this file]**

The same crate list (`gglib-core gglib-db … gglib-agent …`) appears verbatim on lines ~82, ~413, ~622, ~649, ~755, and ~946. Adding `gglib-agent` required six separate spot-edits in this PR, and any future crate will require the same.

YAML doesn't support variable substitution natively, but GitHub Actions `env:` vars can be used for single-line values. The cleanest fix is to define:

```yaml
env:
  ALL_CRATES: "gglib-core gglib-db gglib-gguf gglib-hf gglib-download gglib-mcp gglib-agent gglib-proxy gglib-runtime gglib-gui gglib-cli gglib-axum gglib-tauri gglib-build-info gglib-voice"
```

at the job level and replace every `for CRATE in …` with `for CRATE in $ALL_CRATES`. Not required in this PR, but a follow-up issue / task would prevent future six-edit accidents.

crates/gglib-mcp/src/tool_executor.rs
```diff
@@ -0,0 +1,218 @@
...
+    async fn list_tools(&self) -> Vec<ToolDefinition> {
+        self.mcp
+            .list_all_tools()
+            .await
+            .into_iter()
+            .flat_map(|(server_id, tools)| {
+                tools.into_iter().map(move |t| ToolDefinition {
+                        name: format!("{server_id}:{}", t.name),
+                    description: t.description,
+                    input_schema: t.input_schema,
```
mmogr, line 84: **[Low — inconsistent indentation inside closure]**

```rust
tools.into_iter().map(move |t| ToolDefinition {
        name: format!("{server_id}:{}", t.name),   // ← 8 spaces
    description: t.description,                     // ← 4 spaces
    input_schema: t.input_schema,                   // ← 4 spaces
})
```

`name:` is indented 8 spaces from the opening `{` but `description:` and `input_schema:` are at 4. This is a `rustfmt` violation. Either let `cargo fmt` fix it or align all three fields consistently.

crates/gglib-agent/src/agent_loop.rs
```diff
@@ -0,0 +1,376 @@
...
+        .send(AgentEvent::Error {
+            message: message.to_owned(),
+        })
+        .await;
+}
+
+/// Emit an [`AgentEvent::Error`] and return `Err(`[`AgentError::Internal`]`)`.
+///
+/// Collapses the repeated pattern:
+/// ```text
```
mmogr: **[Info — naming shadows `anyhow::bail!` semantics]**

The name `bail_internal` closely mirrors the `anyhow::bail!` macro convention, which some readers will expect to act like an early-return macro. Here it's a plain `async fn` that must be called with `.await` and whose return value must be forwarded with `return`.

Consider renaming to something that doesn't imply macro behaviour:
- `abort_internal(tx, msg).await` — makes the destructive intent explicit
- `fail_loop(tx, msg).await`

The doc-comment already explains the pattern well; the naming is the only ambiguity.

Cargo.lock
```diff
@@ -1813,6 +1851,20 @@ dependencies = [
  "wasip3",
 ]
 
+[[package]]
+name = "gglib-agent"
+version = "0.5.2"
```
mmogr, line 1856: **[Low — duplicate `nix` and `cfg_aliases` versions introduced by `rustyline`]**

`rustyline 14.0.0` pulls in `nix 0.28.0` and `cfg_aliases 0.1.1`, which conflict with the existing `nix 0.30.1` / `cfg_aliases 0.2.1` already in the tree. Cargo resolves these as two separate compiled copies, adding to binary size and compile time.

`rustyline` is only needed by the CLI REPL. If it is not already behind a `cfg(feature)` in `gglib-cli`, consider adding a `readline` feature flag so server builds (`gglib-axum`, `gglib-gui`) don't pull in the readline stack at all.

This won't get worse if left as-is, but it's worth a note for the next batch of dep cleanups.

crates/gglib-mcp/src/tool_executor.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// returns an error response from the MCP server is represented as a
+    /// `ToolResult { success: false, … }` so the agent can reason about the
+    /// failure and retry or adjust.
+    async fn execute(&self, call: &ToolCall) -> anyhow::Result<ToolResult> {
+        // ---- Resolve server_id and bare tool name from call.name ------------
+        //
+        // Only qualified names are accepted: "{server_id}:{tool_name}".
+        // The `:` separator is unambiguous because MCP tool names
+        // (`[a-zA-Z0-9_-]+`) cannot contain `:`.
+        let (server_id, bare_name): (i64, &str) = if let Some((prefix, bare)) =
```
mmogr: **[Low — unnecessary key/value clone loop when converting `serde_json::Map` → `HashMap`]**

```rust
serde_json::Value::Object(map) => {
    map.iter().map(|(k, v)| (k.clone(), v.clone())).collect()
}
```

`serde_json::Map<String, Value>` implements `IntoIterator<Item = (String, Value)>` and is itself a `BTreeMap`/`IndexMap` wrapper. The clone loop can be replaced with a direct `.clone().into_iter().collect()` — or, since you own `call.arguments` (it's a `&ToolCall` borrow), consider destructuring the arguments by calling `call.arguments.clone()` in one shot:

```rust
serde_json::Value::Object(map) => map.clone().into_iter().collect(),
```

Minor, but keeps noise out of the hot path given this runs on every tool execution.

crates/gglib-runtime/src/ports_impl/llm_completion/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Timeout (seconds) for the `.send()` phase of each LLM request.
+///
+/// Covers TCP connect + TLS handshake + HTTP response headers.  Does **not**
+/// apply to the streaming body, which can take arbitrarily long during prompt
+/// pre-fill.  Chosen conservatively; the llama-server is always local.
+const LLM_CONNECT_TIMEOUT_SECS: u64 = 30;
+use async_trait::async_trait;
+use futures_core::Stream;
+use futures_util::StreamExt as _;
+use reqwest::Client;
```
mmogr: **[Low — constant defined *after* the `use` block that follows it]**

`LLM_CONNECT_TIMEOUT_SECS` is declared on line 29, then `use` statements continue on lines 31-40. Rust allows this but it breaks the conventional order: doc comment → `use` statements → constants → structs → impls. `rustfmt` won't reorder items, but by convention constants should either precede or follow all imports, not be interleaved.

Move the constant (and its doc comment) to after the `use` block, above the `// === Adapter struct ===` section.

mmogr (commented):
All inline comments added. The PR is in good shape — the architecture is clean and the hexagonal boundary is correctly enforced. The actionable items to address before merging (or soon after):

**Must-fix (medium):**
- `gglib-runtime/Cargo.toml`: use `workspace = true` for `futures-core`, `futures-util`, and `reqwest` (those workspace deps were added in this same PR specifically to avoid version drift).

**Should-fix (low):**
- `gglib-mcp/tool_executor.rs`: indentation inconsistency in `list_tools` closure will fail `cargo fmt --check`.
- `context_pruning.rs`: double blank line before `#[cfg(test)]` — same issue.
- `loop_detection.rs`: replace `and_modify().or_insert(1)` with the clearer `or_insert(0); *entry += 1` pattern.
- CLI REPL: construct one `reqwest::Client` at startup, share it across turns via `LlmCompletionAdapter::with_client`.

**Nice-to-have (follow-up issues):**
- Add semaphore-cap test for `execute_tools_parallel`.
- Add `MAX_TOOL_CALL_INDEX`-exceeded test for `stream_collector`.
- Consolidate the 6× duplicated crate list in `badges.yml` into a single `env:` variable.
- Gate `rustyline` behind a `readline` feature flag in `gglib-cli` to avoid compile-time cost in server builds.

PR #320, opened 2026-03-07
fix: pre-built llama.cpp install fails with "Found 0 of 2" on first-run setup

### LINKED ISSUE #319: bug: pre-built llama.cpp install fails with "Found 0 of 2" on first-run setup
## Summary

When a user runs `make setup` without building llama.cpp from source, then launches the GUI and goes through the first-run setup wizard, the installation of pre-built binaries fails immediately after downloading the archive:

```
Installation failed
internal error: Failed to install llama.cpp: Failed to extract all required binaries. Found 0 of 2
```

## Root cause (two bugs, same file)

Both bugs are in `crates/gglib-runtime/src/llama/download/mod.rs`.

---

### Bug 1 — stale `build/bin/` path filter (primary — causes the reported error)

In `extract_binaries_tar_gz` the code filters archive entries with:

```rust
// Binaries live in build/bin/ inside the archive
if !entry_name.contains("build/bin/") {
    continue;
}
```

This path never existed in llama.cpp GitHub release archives. The actual archive structure is:

```
llama-b<tag>/llama-server
llama-b<tag>/llama-cli
llama-b<tag>/libggml-metal.0.dylib
...
```

Every entry is rejected by the filter → `extracted_binaries` stays `0` → the bail fires.

---

### Bug 2 — wrong `bin_dir` target path (secondary — would surface after Bug 1 is fixed)

All three `download_prebuilt_binaries*` functions compute:

```rust
let bin_dir = gglib_dir.join("bin");  // lines 648, 728, 800
```

But `llama_server_path()` (in `gglib-core/src/paths/llama.rs`) expects binaries at:

```
<resource_root>/.llama/bin/llama-server
```

So binaries would be extracted to `<data_root>/bin/` but the post-install verification looks in `<data_root>/.llama/bin/` → "Installation verification failed: binaries not found after extraction".

---

### Bug 3 — `save_prebuilt_config` writes to the wrong path (tertiary)

`save_prebuilt_config` writes `llama-config.json` directly under `data_root()` but `llama_config_path()` expects it at `resource_root()/.llama/llama-config.json`.

## Fix

1. Replace the `"build/bin/"` filter in `extract_binaries_tar_gz` with a path-component count check so it accepts files exactly one level deep (`<dir>/<file>`), which matches the modern archive layout and is resilient to tag name changes.
2. Change `gglib_dir.join("bin")` → `gglib_dir.join(".llama").join("bin")` in all three download functions.
3. Change the config write path in `save_prebuilt_config` to `gglib_dir.join(".llama").join("llama-config.json")`.
4. Add a unit test that creates a synthetic modern-layout tar.gz archive and verifies `extract_binaries_tar_gz` succeeds and produces both required binaries.

## Affected platforms

All platforms (macOS, Linux, Windows) since the path filter is not platform-gated.

## Steps to reproduce

1. `make setup` — choose **not** to build llama.cpp from source
2. Launch the GUI
3. Go through the first-run setup wizard
4. Watch the install step fail with the message above

mmogr:
Fixes #319

## What was broken

When a user ran `make setup` (without building llama.cpp from source) and then launched the GUI, the first-run setup wizard would download the pre-built archive but immediately fail with:

```
Installation failed
internal error: Failed to install llama.cpp: Failed to extract all required binaries. Found 0 of 2
```

Three bugs in `crates/gglib-runtime/src/llama/download/mod.rs` caused this cascade.

---

### 1. `fix(download): replace stale build/bin/ filter with component-count guard`

The extraction loop for `.tar.gz` archives filtered entries with:
```rust
if !entry_name.contains("build/bin/") { continue; }
```
This path never existed in llama.cpp release archives. The actual structure is:
```
llama-b<tag>/llama-server
llama-b<tag>/llama-cli
llama-b<tag>/libggml-metal.0.dylib
```
Every entry was rejected → `extracted_binaries` stayed `0` → the error fired.

**Fix**: replaced with a path-components length check — skip entries that are not exactly one level deep inside the archive's top-level directory. Resilient to future tag name changes.

### 2. `fix(download): extract pre-built binaries into .llama/bin/ to match expected paths`

All three `download_prebuilt_binaries*` functions used:
```rust
let bin_dir = gglib_dir.join("bin");
```
But `llama_server_path()` / `llama_cli_path()` (in `gglib-core`) expect binaries at `resource_root()/.llama/bin/`. The binaries would have landed in the wrong directory and the post-install verification step would have failed:
> "Installation verification failed: binaries not found after extraction"

**Fix**: changed to `gglib_dir.join(".llama").join("bin")` in all three functions.

### 3. `fix(download): write llama-config.json into .llama/ to match llama_config_path()`

`save_prebuilt_config` wrote the config to `data_root()/llama-config.json` but `llama_config_path()` expects it at `resource_root()/.llama/llama-config.json`.

**Fix**: changed the write path accordingly.

### 4. `test(download): add unit test for modern llama.cpp tar.gz archive layout`

Added a unit test that constructs a synthetic tar.gz archive matching the actual `llama-b<tag>/<filename>` release layout, calls `extract_binaries_tar_gz`, and asserts both required binaries and a shared library are properly extracted. This test would have caught the regression in commit 1.

---

## Testing

```
cargo test --package gglib-runtime --features prebuilt
```

New test: `llama::download::tests::test_extract_binaries_tar_gz_modern_layout` ✓

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 
 # llama.cpp local installation
 /.llama
+/downloads
 
 # Web UI build output
 /web_ui
--- a/crates/gglib-runtime/src/llama/download/mod.rs
+++ b/crates/gglib-runtime/src/llama/download/mod.rs
@@ -935,8 +935,7 @@ mod tests {
         // including a symlink entry whose target is not in the archive (dangling).
         // Real macOS llama.cpp releases contain such versioned-dylib symlinks.
         {
-            let archive_file =
-                File::create(&archive_path).expect("failed to create archive file");
+            let archive_file = File::create(&archive_path).expect("failed to create archive file");
             let gz = GzEncoder::new(archive_file, Compression::fast());
             let mut tar = Builder::new(gz);
 
@@ -968,9 +967,7 @@ mod tests {
             link_header.set_entry_type(tar::EntryType::Symlink);
             link_header.set_size(0);
             link_header.set_mode(0o777);
-            link_header
-                .set_link_name("libggml-metal.0.dylib")
-                .unwrap();
+            link_header.set_link_name("libggml-metal.0.dylib").unwrap();
             link_header.set_cksum();
             tar.append_data(&mut link_header, "llama-b9999/libggml.dylib", &b""[..])
                 .unwrap();
--- a/crates/gglib-runtime/src/llama/download/mod.rs
+++ b/crates/gglib-runtime/src/llama/download/mod.rs
@@ -647,14 +647,14 @@ pub async fn download_prebuilt_binaries() -> Result<()> {
     let gglib_dir = path_err(data_root())?;
     let download_dir = gglib_dir.join("downloads");
     let zip_path = download_dir.join(&asset.name);
-    let bin_dir = gglib_dir.join("bin");
+    let bin_dir = gglib_dir.join(".llama").join("bin");
 
     // Download the archive
     download_with_progress(&client, &asset.browser_download_url, &zip_path).await?;
     println!();
 
     // Extract binaries
-    extract_binaries(&zip_path, &bin_dir)?;
+    extract_binaries(&zip_path, &bin_dir)?
 
     // Windows: Also download CUDA runtime DLLs
     #[cfg(target_os = "windows")]
@@ -727,7 +727,7 @@ pub async fn download_prebuilt_binaries_with_callback(
     let gglib_dir = path_err(data_root())?;
     let download_dir = gglib_dir.join("downloads");
     let zip_path = download_dir.join(&asset.name);
-    let bin_dir = gglib_dir.join("bin");
+    let bin_dir = gglib_dir.join(".llama").join("bin");
 
     // Download the archive
     if let Some(callback) = progress_callback {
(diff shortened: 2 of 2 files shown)


mmogr:
**Round 2 fix pushed** (commits `b428eca`, `9e2473a`)

After the round-1 fixes resolved the `Found 0 of 2` error, testing revealed a second crash:

```
Installation failed
internal error: Failed to install llama.cpp: No such file or directory (os error 2)
```

**Root cause**: Modern macOS llama.cpp release archives contain versioned dylib **symlink entries** (e.g. `libggml.dylib → libggml-metal.0.dylib`). After `entry.unpack()` creates the symlink in `bin_dir`, the `#[cfg(unix)]` chmod block called `fs::metadata(&dest_path)?` — which follows symlinks (`stat`, not `lstat`). If the symlink target hadn't been extracted yet, this returned `ENOENT` with no context, propagating all the way up as the bare error shown above.

**Fix**:
- Both `extract_binaries_tar_gz` and `extract_binaries_zip`: replace `fs::metadata` with `fs::symlink_metadata` (lstat — doesn't follow symlinks), then skip `set_permissions` entirely for symlink entries (you can't `chmod` a symlink on macOS/Linux anyway)
- Add `.with_context()` to all previously bare `fs::` error paths in the extraction and config-write code
- Unit test extended to include a dangling symlink entry that would have triggered the crash before the fix

PR #526, opened 2026-05-31
fix(download): fix download speed/ETA accuracy (sliding window estimator + XetPoller scan targets)

### LINKED ISSUE #525: fix: download speed and ETA displays wildly fluctuate due to broken EWA smoothing
## Bug Report

**Symptom:** The download speed and ETA indicators in both the GUI and CLI fluctuate wildly and appear inaccurate, even when the actual download speed (as seen in the system monitor) is completely stable.

## Root Cause

The speed calculation uses an Exponentially Weighted Average (EWA) with a smoothing factor `α = 0.02`. At 250ms tick intervals this gives a **~34-tick half-life ≈ 8.5 seconds** for convergence. The displayed value always lags far behind the true speed — especially at start-up when it initialises from `0.0` — which means it perpetually "chases" the real value without ever settling.

There are two independent copies of this broken logic:

| Path | File | Issue |
|---|---|---|
| GUI SSE bridge | `crates/gglib-download/src/manager/mod.rs` L56 | `EWA_SMOOTHING = 0.02`, ~8.5s convergence lag |
| Non-TTY CLI (`PlainProgress`) | `crates/gglib-download/src/cli_exec/exec/progress.rs` L21 | Same `EWA_SMOOTHING = 0.02` |

Additionally, `indicatif`'s internal `{bytes_per_sec}` and `{eta}` placeholders (used in the TTY CLI and `FancyProgress` fast-download path) use their own internal estimation algorithm, producing different numbers than the GUI — breaking GUI/CLI parity.

## Planned Fix

Replace all EWA-based speed calculations with a shared **8-second sliding window estimator**, implemented as a new `SlidingWindowRate` struct in `crates/gglib-download/src/progress/rate.rs`. Speed = `(newest_bytes − oldest_bytes) / (newest_time − oldest_time)` over the last 8 seconds of samples — accurate, no tuning constants, matches system monitor behaviour.

**Architecture:**
- **DRY:** Single `SlidingWindowRate` struct + standalone `format_eta()` utility, used by all display paths
- **GUI/CLI parity:** Replace `indicatif`'s `{bytes_per_sec}` and `{eta}` template placeholders with `{prefix}` (a free string slot), injecting our own computed `"X.XX MB/s  eta MM:SS"` string via `bar.set_prefix()` — completely bypassing `indicatif`'s internal estimator
- **100% coverage:** All four display paths unified — GUI SSE bridge, TTY CLI (`CliDownloadEventEmitter`), non-TTY CLI (`PlainProgress`), and fast-download path (`FancyProgress`)
- **ETA parity:** `format_eta()` is a shared standalone function producing consistent formatting across all paths; `indicatif`'s `{eta}` is removed from templates entirely

**Files changed:**
- `crates/gglib-download/src/progress/rate.rs` — **new file** (`SlidingWindowRate` + `format_eta` + unit tests)
- `crates/gglib-download/src/progress/mod.rs` — re-export new types
- `crates/gglib-download/src/manager/mod.rs` — use `SlidingWindowRate`; remove `EWA_SMOOTHING`
- `crates/gglib-download/src/cli_emitter.rs` — swap `{bytes_per_sec}`+`{eta}` → `{prefix}`
- `crates/gglib-download/src/cli_exec/exec/progress.rs` — `FancyProgress` + `PlainProgress` both updated

**No changes to:** public API surface, `DownloadEvent` type shape, SSE serialisation, or any frontend code.

## Research

- **`indicatif` 0.18:** Uses a time-based double-smoothed EWA with 15-second window internally. Good, but not injectable from outside.
- **`tqdm`:** EMA with α=0.3 (15× more responsive than our 0.02, but still a tuning constant).
- **`wget`/`curl`/`aria2`:** All use rolling window averages over a recent time window — the industry standard for download speed displays.

Fixes the speed/ETA display for all download types.

mmogr:
Closes #525

## Problem

Download speed and ETA indicators in both the GUI and CLI fluctuated wildly. After implementing the sliding window estimator (see below), the symptom persisted: system monitor showed a stable 90 MB/s, but gglib displayed values that jumped from 4 → 250 → 50 → 1000 → 30 MB/s.

Two separate bugs were found and fixed.

---

### Bug 1 — Slow/wrong estimator algorithm

Both the GUI bridge task and the CLI progress bars used an exponentially-weighted average with α=0.02. At a 250 ms tick interval that gives a ~34-tick half-life (~8.5 s convergence lag), and speed initialises at 0.0 so it takes ~28 s to reach 90% of the true value. Two independent copies of this bug existed in the codebase.

A second issue: indicatif's `{bytes_per_sec}` and `{eta}` placeholders are computed internally by indicatif from `set_position()` calls and cannot be overridden — so even if the bridge task sent correct speed/ETA values, the CLI bars would still show indicatif's own fluctuating estimate.

### Bug 2 — XetPoller scanning the wrong filesystem locations (root cause of spikes)

When `hf_xet` is active, `huggingface_hub` bypasses `tqdm.update()` entirely, so the Python subprocess never emits `Progress` events. `XetPoller` compensates by stat-polling a list of target paths and reporting the total bytes on disk every 250 ms.

Two problems with how those targets were chosen:

1. **Atomic rename spike** — targets included `dest/{filename}` (the final output file). When a transfer completes, `huggingface_hub` atomically renames `.incomplete` → final file. That makes the full file size appear in one 250 ms tick: e.g. a 5 GB file produces a momentary `5 GB / 0.25 s = 20 GB/s` reading. The `SlidingWindowRate` then spends several seconds averaging this spike away while displaying wild intermediate values.

2. **Global HF cache not scanned** — `hf_xet` frequently writes blobs to the system-wide hub cache (`~/.cache/huggingface/hub/models--owner--repo/blobs/`) rather than the local `dest/.cache` staging directory. When that path wasn't in the target list, XetPoller saw 0 bytes for the first ~30 s of a download (explaining the "stuck at 4 MB/s" observation), then received the full blob size in one tick when the rename fired.

---

## Fix

**Shared `SlidingWindowRate` struct** (`progress/rate.rs`) — retains timestamped byte-count samples over an 8-second window and reports `Δbytes / Δtime` across that window. This matches what system monitors show, requires no tuning constants, and is accurate from the second sample onward.

**GUI / SSE path** (`manager/mod.rs`) — the bridge task now calls `rate.record()` + `rate.speed_bps()` on every tick. `DownloadEvent::progress()` in `gglib-core` derives `eta_seconds` from `speed_bps`, so GUI ETA is also fixed with no changes to the core crate.

**CLI indicatif bypass** (`cli_emitter.rs`, `cli_exec/exec/progress.rs`) — `{bytes_per_sec}` and `{eta}` slots removed from all templates and replaced with `{prefix}`. Each bar now holds a `SlidingWindowRate`; speed and ETA are formatted as `"X.XX MB/s  eta MM:SS"` and written via `bar.set_prefix()`, completely bypassing indicatif's internal estimator. `set_position()` is still called so `{bytes}` / `{total_bytes}` / `{bar}` continue to work correctly.

**`PlainProgress` (non-TTY)** now also emits ETA — it was previously omitted entirely.

**XetPoller scan targets** (`cli_exec/exec/python_bridge.rs`, `xet_poller.rs`):
- Removed per-file final-path targets (eliminates the rename spike)
- Added the global HF hub blob directory for the repo as a second target, resolving via `HF_HUB_CACHE` → `HUGGINGFACE_HUB_CACHE` → `HF_HOME/hub` → `$HOME/.cache/huggingface/hub` (same priority as Python's `huggingface_hub`); missing paths are silently skipped
- Made the high-water-mark check strictly monotonic: if polled bytes decrease (e.g. mid-rename), `last_bytes` resets to 0 so the next genuine increase fires immediately

---

## Commits

1. `feat(download): add SlidingWindowRate and format_eta to progress module` — new `rate.rs` with full unit test suite
2. `fix(download): replace EWA bridge task speed with SlidingWindowRate` — `manager/mod.rs`; removes `calculate_speed_eta` helper
3. `fix(download): bypass indicatif estimator in CliDownloadEventEmitter` — `cli_emitter.rs`
4. `fix(download): replace EWA in FancyProgress and PlainProgress` — `cli_exec/exec/progress.rs`
5. `docs(download): document SlidingWindowRate and format_eta in progress README`
6. `fix(download): resolve all strict clippy warnings`
7. `fix(download): fix XetPoller scan targets to eliminate speed spikes` — root cause fix for hf-xet downloads

### DIFF
--- a/crates/gglib-download/src/cli_emitter.rs
+++ b/crates/gglib-download/src/cli_emitter.rs
@@ -11,8 +11,11 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::time::Duration;
+use std::time::Instant;
 
 use indicatif::{HumanBytes, MultiProgress, ProgressBar, ProgressStyle};
+
+use crate::progress::{SlidingWindowRate, format_eta};
 
 use gglib_core::download::DownloadEvent;
 use gglib_core::events::AppEvent;
@@ -31,7 +34,7 @@ use gglib_core::ports::{AppEventEmitter, DownloadEventEmitterPort};
 // The unified template renders fine even when the bar's length is unknown
 // (the bar widget appears empty until `set_length` is called with a real
 // total) and avoids the mid-stream style switch entirely.
-const BAR_TEMPLATE: &str = "{spinner:.cyan} {wide_msg} [{bar:30.cyan/blue}] {bytes}/{total_bytes} @ {bytes_per_sec} eta {eta}";
+const BAR_TEMPLATE: &str = "{spinner:.cyan} {wide_msg} [{bar:30.cyan/blue}] {bytes}/{total_bytes} @ {prefix}";
 const TICK_INTERVAL: Duration = Duration::from_millis(120);
 
 // ─── CliDownloadEventEmitter ─────────────────────────────────────────────────
@@ -47,7 +50,7 @@ const TICK_INTERVAL: Duration = Duration::from_millis(120);
 /// can suspend rendering while collecting user input without tearing the display.
 pub struct CliDownloadEventEmitter {
     multi_progress: Arc<MultiProgress>,
-    bars: Mutex<HashMap<String, ProgressBar>>,
+    bars: Mutex<HashMap<String, (ProgressBar, SlidingWindowRate)>>,
 }
 
 impl CliDownloadEventEmitter {
@@ -81,7 +84,7 @@ impl CliDownloadEventEmitter {
     /// first and [`Self::resume_animation`] when done.
     pub fn pause_animation(&self) {
         if let Ok(bars) = self.bars.lock() {
-            for bar in bars.values() {
+            for (bar, _) in bars.values() {
                 bar.disable_steady_tick();
             }
         }
@@ -92,7 +95,7 @@ impl CliDownloadEventEmitter {
     /// Pairs with [`Self::pause_animation`].
     pub fn resume_animation(&self) {
         if let Ok(bars) = self.bars.lock() {
-            for bar in bars.values() {
+            for (bar, _) in bars.values() {
                 bar.enable_steady_tick(TICK_INTERVAL);
             }
         }
@@ -136,7 +139,7 @@ impl DownloadEventEmitterPort for CliDownloadEventEmitter {
                 bar.set_message(label);
 
                 if let Ok(mut bars) = self.bars.lock() {
-                    bars.insert(id, bar);
+                    bars.insert(id, (bar, SlidingWindowRate::new()));
                 }
             }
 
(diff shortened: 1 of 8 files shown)


mmogr (commented):
Added follow-up fix in 04a4e1b to address residual speed spikes reported after the prior root-cause patch.

What changed:
- XetPoller directory scans now count only in-flight `*.incomplete` files
- Synthetic bytes are baseline-relative (`current - startup`) so pre-existing cache artifacts don't pollute current-request progress
- Synthetic downloaded is clamped to expected total when known

Validation:
- `cargo test --package gglib-download` (109 passed)
- `cargo clippy --package gglib-download -- -D warnings`

If speed still fluctuates after this, I can add temporary debug telemetry to log raw poller samples (raw sum, baseline, emitted bytes, and per-tick delta) for one reproducer run.

mmogr (commented):
Small correction: I accidentally opened then deleted a pending review while posting follow-up notes. No code or review state changed; the intended status update is the latest comment plus commit 04a4e1b.

mmogr (commented):
Consolidated summary (author update): residual fluctuation fix landed in commit 04a4e1b, now in branch. Key change is that synthetic progress uses baseline-relative growth of in-flight `.incomplete` files only, rather than absolute cache-directory size.

Issue #558, opened 2026-06-28
Epic: GGLib Home Server (Distributed Compute & Orchestration)


mmogr:
## Overview

GGLib's long-term vision is a **"Brain vs. Muscle"** architecture: a 24/7 low-power **Control Plane** (NAS, Raspberry Pi, or similar) paired with a high-power **Compute Node** (RTX gaming PC or workstation). The Control Plane handles state, databases, memory injection, routing, and always-on availability. The Compute Node handles heavy model inference on demand.

This Epic tracks all engineering work to realize that vision across three distinct phases.

---

## Phase 1 — Opt-in LAN Discovery `v1.2.0`

**Goal:** Harden the default single-machine experience and introduce the foundational networking primitives that Phase 2 will build on.

### Security-first defaults
- `gglib web` currently binds to `0.0.0.0` — this must change to `127.0.0.1` by default
- A new `--share-lan` flag (or persistent setting) opts the user in to LAN exposure
- CORS must tighten to a new `CorsConfig::LocalOnly` mode when not in LAN mode

### mDNS / local service discovery
- When `--share-lan` is active, register `_gglib._tcp.local.` via the `mdns-sd` crate
- This makes the node addressable as `gglib.local` on the LAN without any manual IP configuration
- The mDNS responder deregisters cleanly on shutdown

### Settings persistence
- New SQLite columns `bind_host` and `share_lan` persist the user's preference across restarts
- CLI flags override stored settings; stored settings override compiled-in defaults

**Milestone:** [v1.2.0 - Local Discovery](https://github.com/mmogr/gglib/milestone/2)

---

## Phase 2 — Docker Control Plane `v1.3.0`

**Goal:** Package the always-on, stateful half of GGLib as a first-class Docker container suitable for running on a NAS (Synology, TrueNAS, Unraid) or Raspberry Pi 5.

### Control Plane container
- Headless axum API server with no llama-server dependency
- Owns the SQLite database, model catalog, download manager, chat history
- Exposes a stable internal RPC / REST API that Compute Nodes connect to

### Remote Compute Node pairing
- Compute Nodes (running the full `gglib` binary with llama-server) register themselves with the Control Plane via the mDNS service from Phase 1
- The Control Plane routes inference requests to available Compute Nodes based on model availability and load

### OpenWebUI integration hardening
- Control Plane serves a clean OpenAI-compatible `/v1/*` API
[...]

PR #629, opened 2026-07-23
Secure default localhost bind and CORS configuration

### LINKED ISSUE #559: Phase 1a: Implement CorsConfig::LocalOnly and restrict default web binding to 127.0.0.1
## Goal

Fix the security regression where `gglib web` binds to `0.0.0.0` by default, and introduce a localhost-only CORS mode.

## Context

See parent epic #558 and the Phase 1 architecture plan.

Currently `start_server()` in `crates/gglib-axum/src/bootstrap.rs` hardcodes `0.0.0.0:{port}` regardless of any host setting. `gglib proxy` is already correct (it accepts a `--host` flag defaulting to `127.0.0.1` via `ProxySupervisor`), but `gglib web` is not.

### `crates/gglib-axum/src/bootstrap.rs`
- Add `pub host: String` field to `ServerConfig` struct
- Set default to `"127.0.0.1".to_string()` in `with_defaults()`
- Add new `CorsConfig::LocalOnly` variant to the `CorsConfig` enum
- Fix `start_server()`: replace `format!("0.0.0.0:{}", config.port)` → `format!("{}:{}", config.host, config.port)`

### `crates/gglib-axum/src/routes.rs`
- Add match arm in `build_cors_layer()` for `CorsConfig::LocalOnly`
- Use `AllowOrigin::predicate` to accept only origins matching `localhost` or `127.0.0.1`

### `crates/gglib-cli/src/handlers/web.rs`
- Change default `cors: CorsConfig::AllowAll` → `CorsConfig::LocalOnly`

## Acceptance Criteria
- [ ] `cargo run -- web --api-only --port 8080` binds only to `127.0.0.1:8080` (verified with `lsof -i :8080`)
- [ ] Cross-origin requests from non-localhost origins are rejected with 403
- [ ] All existing tests pass
- [ ] Tauri embedded server (already binds to `127.0.0.1:0`) is unaffected

mmogr:
## Goal

Fixes #559
Resolves a security regression where `gglib web` binds to `0.0.0.0` by default, exposing the API to the local network. This PR secures the local development environment by enforcing a `127.0.0.1` default bind, providing a `--host` CLI escape hatch for intentional LAN testing, and unifying a strict localhost-only CORS policy across both the WebUI and Proxy servers.

## Context

See parent epic #558 and the Phase 1 architecture plan.

Previously, the Axum bootstrap hardcoded `0.0.0.0:{port}` and utilized a permissive CORS policy. Additionally, there was a parity gap: the Proxy server defaulted to `127.0.0.1` but still used a permissive `allow_origin(Any)` layer. To prevent cross-origin attacks from malicious sites, this PR centralizes our security posture to ensure both servers symmetrically reject non-localhost origins.

## Key Changes

* **Secure Network Bind:** Added a `with_host` builder to `ServerConfig` to support dynamic host binding. The `gglib web` command now defaults to `127.0.0.1` (configurable via `--host`), and the startup banner dynamically reflects the actual bind address.
* **Shared Security Utility:** Extracted `CorsConfig` and a new `is_local_origin` utility into `gglib-core`. The utility uses secure native URL parsing (via the `url` crate) to strictly reject userinfo injections (e.g., `http://user@localhost`).
* **Proxy Parity:** Both the Axum web server and the Proxy now use the shared `is_local_origin` logic for their CORS layers, fully respecting the `CorsConfig::LocalOnly` default. 
* **Tauri Compatibility:** Added explicit guards to `is_local_origin` to ensure Tauri's desktop schemes (`tauri://localhost`, `asset://localhost`) and standard webview origins (`http://tauri.localhost`) bypass strict web CORS restrictions natively.
* **Documentation & Testing:** Updated `README.md` and crate docs to reflect the new `127.0.0.1` defaults, and added comprehensive integration tests validating the `LocalOnly` CORS policy across both Axum and the Proxy.

## Acceptance Criteria

* [ ] `cargo run -- web --api-only --port 8080` binds **only** to `127.0.0.1:8080` (verified via `lsof -i :8080`).
* [ ] The Proxy server mirrors this security posture, strictly rejecting non-localhost origins by default.
* [ ] Cross-origin requests from non-localhost origins (e.g., `http://evil.example.com`) are rejected with a 403 or missing `Access-Control-Allow-Origin` header.
* [ ] Tauri embedded server origins (`tauri://localhost`, `asset://localhost`, `http://tauri.localhost`) are parsed correctly and bypass CORS restrictions.
* [ ] All existing tests pass, and the `check_boundaries.sh` script remains green.

### DIFF
--- a/crates/gglib-axum/src/bootstrap.rs
+++ b/crates/gglib-axum/src/bootstrap.rs
@@ -101,6 +101,13 @@ impl ServerConfig {
         self
     }
 
+    /// Set the bind host (e.g. `127.0.0.1` for localhost-only).
+    #[must_use]
+    pub fn with_host(mut self, host: impl Into<String>) -> Self {
+        self.host = host.into();
+        self
+    }
+
     /// Set CORS to allow specific origins.
     #[must_use]
     pub fn with_allowed_origins(mut self, origins: Vec<String>) -> Self {
--- a/crates/gglib-axum/src/bootstrap.rs
+++ b/crates/gglib-axum/src/bootstrap.rs
@@ -46,15 +46,19 @@ use gglib_core::paths::{
 #[derive(Debug, Clone, Default)]
 pub enum CorsConfig {
     /// Allow all origins (development mode).
-    #[default]
     AllowAll,
     /// Allow specific origins (production mode).
     AllowOrigins(Vec<String>),
+    /// Restrict to local-only access (localhost, 127.0.0.1, ::1).
+    #[default]
+    LocalOnly,
 }
 
 /// Server configuration for the Axum adapter.
 #[derive(Debug, Clone)]
 pub struct ServerConfig {
+    /// Host to bind the HTTP server.
+    pub host: String,
     /// Port for the HTTP server.
     pub port: u16,
     /// Base port for llama-server instances.
@@ -79,6 +83,7 @@ impl ServerConfig {
     /// Create config with default paths.
     pub fn with_defaults() -> Result<Self> {
         Ok(Self {
+            host: "127.0.0.1".into(),
             port: 9887,
             base_port: 9000,
             llama_server_path: llama_server_path()?,
@@ -414,7 +419,7 @@ pub async fn start_server(config: ServerConfig) -> Result<()> {
         crate::routes::create_router(ctx, &config.cors)
     };
 
-    let addr = format!("0.0.0.0:{}", config.port);
+    let addr = format!("{}:{}", config.host, config.port);
     let listener = TcpListener::bind(&addr).await?;
 
     if config.static_dir.is_some() {
--- a/crates/gglib-axum/src/bootstrap.rs
+++ b/crates/gglib-axum/src/bootstrap.rs
@@ -41,18 +41,7 @@ use crate::sse::SseBroadcaster;
 use gglib_core::paths::{
     data_root, database_path, llama_server_path, resolve_models_dir, resource_root,
 };
-
-/// CORS configuration for the web server.
-#[derive(Debug, Clone, Default)]
-pub enum CorsConfig {
-    /// Allow all origins (development mode).
-    AllowAll,
-    /// Allow specific origins (production mode).
-    AllowOrigins(Vec<String>),
-    /// Restrict to local-only access (localhost, 127.0.0.1, ::1).
-    #[default]
-    LocalOnly,
-}
+use gglib_core::CorsConfig;
 
 /// Server configuration for the Axum adapter.
 #[derive(Debug, Clone)]
--- a/crates/gglib-axum/src/bootstrap.rs
+++ b/crates/gglib-axum/src/bootstrap.rs
@@ -38,10 +38,10 @@ use gglib_runtime::system::DefaultSystemProbe;
 use crate::sse::SseBroadcaster;
 
 // Path utilities from core
+use gglib_core::CorsConfig;
 use gglib_core::paths::{
     data_root, database_path, llama_server_path, resolve_models_dir, resource_root,
 };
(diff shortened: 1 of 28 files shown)


crates/gglib-axum/src/bootstrap.rs
```diff
@@ -46,15 +46,19 @@ use gglib_core::paths::{
...
 pub enum CorsConfig {
     /// Allow all origins (development mode).
-    #[default]
     AllowAll,
     /// Allow specific origins (production mode).
     AllowOrigins(Vec<String>),
+    /// Restrict to local-only access (localhost, 127.0.0.1, ::1).
+    #[default]
+    LocalOnly,
 }
```
mmogr: **Design Flaw: Core Domain Leak & Code Duplication**

Defining `CorsConfig` directly inside the `gglib-axum` adapter prevents `gglib-proxy` from reusing the same configuration schema. We should extract this enum to `crates/gglib-core/src/cors.rs` and re-export it here so both HTTP servers can securely share the exact same CORS policy types without breaking existing callers.

```suggestion
pub use gglib_core::CorsConfig;
```

crates/gglib-core/src/is_local_origin.rs
```diff
@@ -0,0 +1,114 @@
...
+    let Some(host_str) = parsed.host_str() else {
+        return false;
+    };
+
+    // `url::Url::host_str()` returns IPv6 addresses with brackets (e.g. `[::1]`),
+    // so strip them for comparison.
+    let host = host_str.trim_start_matches('[').trim_end_matches(']');
+
+    matches!(host, "localhost" | "127.0.0.1" | "::1")
+}
```
mmogr, line 61: **Design Flaw: Missing Tauri Schemes & Userinfo Security Guard**

1. Custom desktop schemes used by Tauri (`tauri://localhost` or `asset://localhost`) are rejected because this function currently restricts schemes strictly to `http` and `https`.
2. Userinfo credentials (e.g., `http://user@localhost`) pass the host check because `Url::parse` resolves the host to `localhost` while preserving `user`.

I'm updating this function to handle Tauri custom schemes prior to parsing, and strictly rejecting any origin that includes userinfo credentials via `parsed.username()` and `parsed.password()`.

```suggestion
pub fn is_local_origin(origin: &str) -> bool {
    // Handle Tauri custom schemes prior to URL parsing
    if let Some(stripped) = origin.strip_prefix("tauri://").or_else(|| origin.strip_prefix("asset://")) {
        return matches!(stripped.trim_end_matches('/'), "localhost" | "127.0.0.1" | "[::1]" | "::1");
    }

    let Ok(parsed) = url::Url::parse(origin) else {
        return false;
    };

    // Only allow http/https schemes for standard URLs
    if !matches!(parsed.scheme(), "http" | "https") {
        return false;
    }

    // Strictly reject userinfo credentials (e.g., http://user@localhost)
    if !parsed.username().is_empty() || parsed.password().is_some() {
        return false;
    }

    let Some(host_str) = parsed.host_str() else {
        return false;
    };

    let host = host_str.trim_start_matches('[').trim_end_matches(']');

    matches!(host, "localhost" | "127.0.0.1" | "::1" | "tauri.localhost")
}
```

crates/gglib-axum/src/bootstrap.rs
mmogr: Fixed in 0cfa33ff — extracted CorsConfig to gglib-core so it can be shared cleanly.

crates/gglib-core/src/is_local_origin.rs
mmogr: Fixed in eab35d50 — updated to handle Tauri custom schemes and reject userinfo injections via the url crate API.

PR #636, opened 2026-07-26
feat(serve/proxy): parity — unified config pipeline & serve/proxy convergence (#630)

### LINKED ISSUE #630: Unify `gglib serve` and `gglib proxy` into a Single Execution Path
## Problem Statement

The `gglib proxy` stack provides essential observability—including real-time prompt ingest progress, KV cache hits, RAM budget health, and slot utilization—via its dashboard. 

However, certain API clients (like VS Code Copilot's BYOK endpoint) require a single-model, pinned API endpoint and cannot dynamically switch models via `/v1/models`. Currently, this use-case is supported via `gglib serve <model>`, but **`serve` bypasses the proxy stack entirely**, utilizing a completely different code path. 

Because of this bypass, requests routed through `serve` lack:
- **Caching** — no KV slot persistence, no RAM cache auto-sizing, no cache reuse tuning.
- **Observability** — no `/v1/proxy/status` endpoints or SSE streams.
- **Security Defaults** — missing `--host 127.0.0.1`, `--metrics`, and `--parallel 1`.
- **Request Normalization** — no response stream normalization or upstream health monitoring.

This architectural divergence has created a severe DRY violation, resulting in two parallel command builders and configuration structs (`ProxyConfig` vs `ServerConfigOptions`) that are actively drifting apart in feature parity.

### Two Command Builders, Two Paths

| Layer | `gglib serve` | `gglib proxy` (and GUI) |
|---|---|---|
| **Command builder** | invocation.rs (simple, CLI-oriented) | server_config.rs → command.rs (canonical) |
| **Process management** | Raw `cmd.status()` (serve.rs) | manager.rs (auto-start, auto-swap, health-check) |
| **HTTP layer** | None (direct llama-server, inherited stdio) | Axum proxy → forward to llama-server (server.rs) |
| **KV cache persistence** | ❌ | ✅ cache_lifecycle.rs (restore→forward→save) |
| **RAM cache auto-sizing** | ❌ | ✅ `manager.rs:490-500` |
| **Dashboard** | ❌ | ✅ dashboard.rs + SSE publisher |
| **`--host 127.0.0.1`** | ❌ (missing — security risk, see #559 / PR #629) | ✅ command.rs |
| **`--metrics`** | ❌ | ✅ command.rs |
| **`--parallel 1`** | ❌ | ✅ command.rs |

### Flag Drift Evidence

The canonical `command.rs` builds flags that `serve` completely misses, and the config structs themselves have drifted:

```rust
// command.rs:167-220 (excerpt)
cmd.arg("--host").arg("127.0.0.1")     // ← serve is MISSING this (security vulnerability)
   .arg("--metrics")                   // ← serve is MISSING this
   .arg("--parallel").arg("1");        // ← serve is MISSING this

// Cache flags (all MISSING from serve):
if let Some(ref slot_path) = config.slot_save_path {
    cmd.arg("--slot-save-path").arg(slot_path);
}
// Furthermore, `mlock` is missing from `build_server_config` entirely.
```

## Architectural Rules

1. **No Backwards Compatibility.** Legacy flags will be dropped. Do not attempt to maintain parity with the old `serve` arguments if they conflict with the unified proxy.
2. **Strict DRY Core.** `gglib serve <model>` must become a specialized, pinned single-model invocation of the unified proxy stack, NOT a bare-metal `llama-server` bypass.
3. **Configuration Cascade.** A single `UnifiedServerConfig` must apply settings in this exact order:
   - **Tier 1:** Explicit Overrides (CLI flags, GUI settings).
   - **Tier 2:** Curated Model Defaults (from model tags, GGUF metadata).
   - **Tier 3:** Global Defaults (e.g., `--host 127.0.0.1`, `--parallel 1`, `--metrics`).
4. **GUI Parity.** The CLI, Axum WebUI, and Tauri app must all consume this exact same engine. 

## Target Architecture

```
gglib serve <model>
    │
    ▼
[Unified Config Resolver]  ← merges ProxyConfig + ServerConfigOptions
    │
    ▼
build_server_config()  ← crates/gglib-runtime/src/server_config.rs:56
    │  
    ▼
ProcessManager::Pinned  ← crates/gglib-runtime/src/process/manager.rs
    │  (Rejects foreign model requests, bypasses auto-swap, RAM auto-sizing)
    │
    ▼
gglib-proxy serve()  ← crates/gglib-proxy/src/server.rs:127
    │  (Axum HTTP layer, request normalization, SSE streaming)
    │
    ▼
[Dashboard]  ← DashboardSnapshot + SSE publisher
[Cache Lifecycle]  ← restore→forward→save
```

## Files to Delete / Refactor

| File | Change |
|---|---|
| `invocation.rs` | **DELETE.** `LlamaCommandBuilder` is the root of the DRY violation. |
| `serve.rs` | **REWRITE.** Remove raw `cmd.status()` patterns; initialize the proxy with a pinned model. |
| `process_runner.rs` | Add missing fields (e.g., `mlock`) to `ServerConfigOptions`. |
| `server_config.rs` | Introduce `UnifiedServerConfig`. Implement the strict 3-tier cascade. |
| `command.rs` | Route `serve` through `build_and_spawn`. Wire `--mlock`. |
| `manager.rs` | Add `ProcessStrategy::Pinned { model_id, model_name }` variant. |
| `servers.rs` (GUI) | Route `ServerOps` through the proxy instead of direct runner starts. |

## Implementation Phases (Sub-Issues)

Convert the following items into individual tracking issues.

- [x] #631 (Phase 1: Foundation — Config & State)
  * **Goal:** Wire missing configuration state through the canonical pipeline.
  * **Tasks:**
    * Add missing fields (e.g., `mlock: bool` / `Option<bool>`) to `StartServerRequest`, `ServerConfigOptions`, and `ServerConfig`.
    * Wire `--mlock` flag emission into `build_command()` with unit tests.

- [x] #632 (Phase 2: Pinned Mode & Serve Rewrite)
  * **Goal:** Build the single-model proxy mode and route `serve` through it.
  * **Tasks:**
    * Create `UnifiedServerConfig` module to merge `ProxyConfig` and `ServerConfigOptions`.
    * Formalize the 3-tier configuration cascade via `resolve_unified_config()`.
    * Route `ServerOps` through `ProxyOps` in Axum (`bootstrap.rs`) and Tauri (`bootstrap.rs`).
    * Add `ProcessStrategy::Pinned { model_id, model_name }` to `manager.rs`. (Behaves like `SingleSwap` but rejects foreign model requests).
    * Rewrite `serve.rs` handler to build `UnifiedServerConfig`, initialize the `Pinned` manager, and start the proxy.
    * Deprecate and delete `LlamaCommandBuilder` completely.

- [ ] #633 (Phase 3: Catalog Filtering, CLI Parity, CI & Docs)
  * **Goal:** Complete the pinned-mode contract, clean up the UX, and ensure regressions are caught.
  * **Tasks:**
    * **Filter `/v1/models` to only return the pinned model so BYOK clients don't see the full catalog.** *(Deferred from Phase 2 — execution is already locked to the pinned model, but the catalog endpoint still advertises every model, so a BYOK client can select one and get a `PinnedModelMismatch` rejection instead of never being offered it.)*
    * Extract shared `CacheArgs` and unify `SamplingArgs` so `serve` and `proxy` parse identically.
    * Update CLI help strings for `gglib serve` to explicitly document the active dashboard (`/v1/proxy/status`) and caching behaviors.
    * Add CI integration tests verifying `/v1/proxy/status` and KV cache persistence work correctly when running `gglib serve <model>`.
    * Update README.md to reflect feature parity.

## Risks & Mitigations

| Risk | Mitigation |
|---|---|
| Startup latency (~1-3s overhead) | Acceptable tradeoff for feature parity; health check gates readiness. |
| Memory overhead (~20-50MB) | Proxy runtime brings required features (caching, SSE) that `serve` users currently lack. |
| `--parallel 1` limits throughput | This is the correct global default for local single-model serving; configurable via overrides. |
| Breaking legacy scripts parsing `serve` stdout | Intentional per Rule 1 (No backwards compatibility). |

### LINKED ISSUE #631: Phase 1: The Foundation (Config & State)
# Phase 1: Foundation — Config & State (mlock plumbing)

**Parent Epic:** [#630](https://github.com/mmogr/gglib/issues/630) — Unify `gglib serve` and `gglib proxy` into a Single Execution Path

## Objective

Wire the missing `mlock` field through the canonical config pipeline so that the GUI frontend's `mlock` setting actually reaches `llama-server` via `--mlock`. This is purely internal plumbing — **no user-visible behavior changes**.

## Problem

`mlock` exists in:
- The CLI path (`LlamaCommandBuilder` in `invocation.rs`) — ✅ works
- The frontend (`ServeConfig` in `index.ts`, mapped in `mappers.ts`) — ✅ sends `mlock: config.mlock ?? false`

But it is **completely absent** from the canonical GUI/proxy pipeline:

| Struct | Location | Has `mlock`? |
|---|---|---|
| `StartServerRequest` (Rust) | `types.rs:338` | ❌ **silently drops frontend `mlock`** |
| `ServerConfigOptions` | `server_config.rs:80` | ❌ |
| `ServerConfig` | `process_runner.rs:20` | ❌ |
| `build_command()` | `command.rs:166` | ❌ no `--mlock` emission |

The frontend sends `mlock` in JSON, but because `StartServerRequest` lacks the field (and lacks `#[serde(deny_unknown_fields)]`), serde silently drops it. Even if it weren't dropped, the downstream structs and command builder don't know about it.

### Step 1: Add `mlock` to Rust `StartServerRequest`

**File:** `crates/gglib-app-services/src/types.rs`

Add field to `StartServerRequest` (line ~338):

```rust
pub struct StartServerRequest {
    pub context_length: Option<u64>,
    pub port: Option<u16>,
    pub jinja: Option<bool>,
    pub reasoning_format: Option<String>,
    pub mtp_draft_n_max: Option<u32>,
    pub mtp_draft_p_min: Option<f32>,
    pub inference_params: Option<gglib_core::domain::InferenceConfig>,
    #[serde(default)]
    pub mlock: bool,  // NEW: from frontend ServeConfig.mlock
}
```

**Test:** Add a serde deserialization test in the existing `update_model_request_tests` module (or a new `start_server_request_tests` module) verifying that `{"mlock": true}` deserializes correctly:

```rust
mod start_server_request_tests {
    use super::*;

    #[test]
    fn mlock_deserializes_from_json() {
        let json = r#"{"mlock": true}"#;
        let req: StartServerRequest = serde_json::from_str(json).unwrap();
        assert!(req.mlock);
    }

    #[test]
    fn mlock_defaults_to_false_when_omitted() {
        let json = r#"{}"#;
        let req: StartServerRequest = serde_json::from_str(json).unwrap();
        assert!(!req.mlock);
    }
}
```

### Step 2: Add `mlock` to `ServerConfigOptions`

**File:** `crates/gglib-core/src/server_config.rs`

Add field to `ServerConfigOptions` (line ~80), after `inference_params`:

```rust
pub struct ServerConfigOptions {
    pub context_size: Option<u64>,
    pub model_server_ctx: Option<usize>,
    pub global_default_ctx: Option<u64>,
    pub port: Option<u16>,
    pub jinja: Option<bool>,
    pub reasoning_format: Option<String>,
    pub mtp_draft_n_max: Option<u32>,
    pub mtp_draft_p_min: Option<f32>,
    pub slot_save_path: Option<PathBuf>,
    pub cache_ram_mb: Option<u64>,
    pub cache_reuse: Option<u32>,
    pub cache_type_k: Option<crate::cache_config::KvCacheType>,
    pub cache_type_v: Option<crate::cache_config::KvCacheType>,
    pub inference_params: Option<InferenceConfig>,
    /// Whether to lock the model in RAM (`--mlock`).
    /// `Some(true)` → enable. `Some(false)` → disable. `None` → default (false).
    pub mlock: Option<bool>,
}
```

### Step 3: Add `mlock` to `ServerConfig` + Builder Method

**File:** `crates/gglib-core/src/ports/process_runner.rs`

Add field to `ServerConfig` (line ~20), after `cache_type_v`:

```rust
pub struct ServerConfig {
    pub model_id: i64,
    pub model_name: String,
    pub model_path: PathBuf,
    pub port: Option<u16>,
    pub base_port: u16,
    pub context_size: Option<u64>,
    pub gpu_layers: Option<i32>,
    pub jinja: bool,
    pub reasoning_format: Option<String>,
    pub spec_draft_n_max: Option<u32>,
    pub spec_draft_p_min: Option<f32>,
    pub inference_config: Option<InferenceConfig>,
    pub extra_args: Vec<String>,
    pub slot_save_path: Option<PathBuf>,
    pub cache_ram_mb: Option<u64>,
    pub cache_reuse: Option<u32>,
    pub cache_type_k: Option<crate::cache_config::KvCacheType>,
    pub cache_type_v: Option<crate::cache_config::KvCacheType>,
    /// Whether to lock the model in RAM (`--mlock`). Default: `false`.
    pub mlock: bool,
}
```

Update `ServerConfig::new()` to set `mlock: false`:

```rust
pub const fn new(model_id: i64, model_name: String, model_path: PathBuf, base_port: u16) -> Self {
    Self {
        // ... existing fields ...
        mlock: false,
    }
}
```

Add builder method after `with_cache_type_v`:

```rust
/// Enable memory lock (`--mlock`).
pub const fn with_mlock(mut self) -> Self {
    self.mlock = true;
    self
}
```

**⚠️ Breaking change for tests:** The `minimal_config()` helper in `command.rs:336` constructs `ServerConfig` with struct literal syntax. You MUST add `mlock: false` to that literal, or all `command.rs` tests will fail to compile.

### Step 4: Wire `mlock` through `build_server_config()`

**File:** `crates/gglib-runtime/src/server_config.rs`

In `build_server_config()`, add resolution after the MTP block (after the `resolve_mtp_args` call, before the function returns):

```rust
// --- Memory lock -------------------------------------------------------------
let mlock = opts.mlock.unwrap_or(false);
if mlock {
    debug!("enabling --mlock for model");
    config = config.with_mlock();
}
```

### Step 5: Wire `--mlock` flag emission in `build_command()`

**File:** `crates/gglib-runtime/src/command.rs`

In `build_command()`, add flag emission after the MTP block (after the `spec_draft_p_min` block, before inference config args):

```rust
// Memory lock
if config.mlock {
    cmd.arg("--mlock");
}
```

**Tests:** Add two tests in the existing `#[cfg(test)]` module, following the established pattern:

```rust
#[test]
fn mlock_omitted_by_default() {
    let config = minimal_config(); // mlock defaults to false
    let cmd = build_command(Path::new("/fake/llama-server"), &config, 5500);
    let args = args_of(&cmd);
    assert!(!args.contains(&"--mlock".to_string()), "mlock should not be present by default");
}

#[test]
fn mlock_emits_flag_when_enabled() {
    let config = ServerConfig {
        mlock: true,
        ..minimal_config()
    };
    let cmd = build_command(Path::new("/fake/llama-server"), &config, 5500);
    let args = args_of(&cmd);
    assert!(args.contains(&"--mlock".to_string()), "--mlock should be present when enabled");
}
```

Remember to update `minimal_config()` to include `mlock: false`.

### Step 6: Update `ServerOps::build_config()` to pass `mlock`

**File:** `crates/gglib-app-services/src/servers.rs`

In the `build_config()` method (line ~158), add `mlock` to the `ServerConfigOptions` construction:

```rust
let opts = ServerConfigOptions {
    // ... existing fields ...
    mlock: Some(request.mlock),  // from StartServerRequest
};
```

Verify that `request.mlock` is accessible (it is, after Step 1).

---

## Deferred to Phase 2

The following items from the original issue scope are **deferred** because they have no Phase 1 consumer and would trigger dead-code warnings under `clippy -D warnings`:

- `UnifiedServerConfig` struct (no consumer yet)
- `resolve_unified_config()` function (no caller yet)
- `#[deprecated]` on `build_server_config()` (fails CI)
- `ServerOps` → `ProxyOps` delegation (requires widening `ModelRuntimePort` signatures first)
- Bootstrap wiring changes (depends on delegation)

These will be absorbed into the Phase 2 issue (#632).

---

## Verification Gates

After completing all 6 steps:

- [ ] `cargo check --workspace` passes
- [ ] `cargo clippy --workspace -- -D warnings` passes
- [ ] `cargo test -p gglib-core` passes (new `mlock` field in `ServerConfig`/`ServerConfigOptions`)
- [ ] `cargo test -p gglib-runtime` passes (new `--mlock` in `build_command`, new tests)
- [ ] `cargo test -p gglib-app-services` passes (ServerOps passes `mlock`, new `StartServerRequest` serde tests)
- [ ] Verify `--mlock` flag appears in `build_command` test output (`mlock_emits_flag_when_enabled`)
- [ ] Verify `--mlock` is absent by default (`mlock_omitted_by_default`)

---

## Files Summary

| Action | File | Change |
|---|---|---|
| Modify | `crates/gglib-app-services/src/types.rs` | Add `mlock: bool` with `#[serde(default)]` to `StartServerRequest` + serde tests |
| Modify | `crates/gglib-core/src/server_config.rs` | Add `mlock: Option<bool>` to `ServerConfigOptions` |
| Modify | `crates/gglib-core/src/ports/process_runner.rs` | Add `mlock: bool` to `ServerConfig`, add `with_mlock()` |
| Modify | `crates/gglib-runtime/src/server_config.rs` | Wire `mlock` through `build_server_config()` |
| Modify | `crates/gglib-runtime/src/command.rs` | Emit `--mlock` in `build_command()` + 2 new tests + update `minimal_config()` |
| Modify | `crates/gglib-app-services/src/servers.rs` | Pass `mlock` in `ServerConfigOptions` construction |

### LINKED ISSUE #632: Phase 2: The Cutover (Pinned Mode & Serve Rewrite)
# Phase 2: Pinned Mode & Serve Rewrite

**Parent Epic:** [#630](https://github.com/mmogr/gglib/issues/630) — Unify `gglib serve` and `gglib proxy` into a Single Execution Path

**Prerequisite:** [#631](https://github.com/mmogr/gglib/issues/631) — Phase 1: Foundation (mlock plumbing) must be merged first.

## Objective

Rewrite `gglib serve` to consume the unified proxy stack, eliminating the parallel `LlamaCommandBuilder` code path entirely. This phase introduces:

1. **`UnifiedServerConfig`** — a single config struct merging `ProxyConfig` + `ServerConfigOptions` with a strict 3-tier cascade resolver.
2. **`ProcessStrategy::Pinned`** — a new manager mode that behaves like `SingleSwap` but rejects foreign model requests (enforcing single-model pinning).
3. **`ServerOps` → `ProxyOps` delegation** — the GUI bootstraps (Axum + Tauri) route all server starts through the proxy pipeline.
4. **`serve.rs` rewrite** — the CLI `gglib serve` handler initializes the proxy with a pinned model instead of spawning `llama-server` directly.
5. **`LlamaCommandBuilder` deletion** — the root DRY violation is removed.

## Target Architecture

```
gglib serve <model>
    │
    ▼
[UnifiedServerConfig]  ← merges ProxyConfig + ServerConfigOptions + CLI args
    │
    ▼
resolve_unified_config()  ← 3-tier cascade: explicit > model defaults > global defaults
    │
    ▼
ProcessManager::Pinned  ← rejects foreign models, bypasses auto-swap
    │
    ▼
gglib-proxy serve()  ← Axum HTTP layer, request normalization, SSE, dashboard
    │
    ▼
[Dashboard] + [Cache Lifecycle] + [SSE Publisher]
```

---

### Problem

`ProxyConfig` (proxy-level settings) and `ServerConfigOptions` (per-model launch params) are separate bags of options with no enforced cascade. `build_server_config()` handles some cascade logic (context size, jinja, reasoning, MTP) but other fields (`mlock`, `gpu_layers`, `cache_ram_mb`) are direct pass-throughs with no fallback.

### Design

**File to create:** `crates/gglib-core/src/unified_server_config.rs` (new module)

```rust
/// Unified configuration for starting a llama-server instance, whether via
/// the proxy stack or directly.
///
/// Captures both per-model launch parameters (context size, GPU layers,
/// cache settings) and proxy-level settings (host, port, cache persistence).
/// Single input to the unified config resolver.
#[derive(Debug, Clone)]
pub struct UnifiedServerConfig {
    // --- Model identity ---
    pub model_id: i64,
    pub model_name: String,
    pub model_path: PathBuf,
    pub model_tags: Vec<String>,

    // --- Tier 1: Explicit Overrides (CLI flags, GUI settings) ---
    pub explicit_port: Option<u16>,
    pub explicit_context_size: Option<u64>,
    pub explicit_gpu_layers: Option<i32>,
    pub explicit_jinja: Option<bool>,
    pub explicit_reasoning_format: Option<String>,
    pub explicit_mtp_draft_n_max: Option<u32>,
    pub explicit_mtp_draft_p_min: Option<f32>,
    pub explicit_mlock: Option<bool>,
    pub explicit_inference_params: Option<InferenceConfig>,
    pub explicit_cache_ram_mb: Option<u64>,
    pub explicit_cache_reuse: Option<u32>,
    pub explicit_cache_type_k: Option<KvCacheType>,
    pub explicit_cache_type_v: Option<KvCacheType>,
    pub explicit_slot_save_path: Option<PathBuf>,

    // --- Tier 2: Model Defaults (from GGUF metadata / model tags) ---
    pub model_server_ctx: Option<usize>,

    // --- Tier 3: Global Defaults ---
    pub global_host: String,              // default "127.0.0.1"
    pub global_port: u16,                 // default proxy port
    pub global_base_port: u16,            // default llama-server base port
    pub global_default_ctx: Option<u64>,
    pub global_cache_enabled: bool,
    pub global_slot_dir: Option<PathBuf>,
    pub global_disk_budget: DiskBudget,
    pub global_inference_override: Option<InferenceConfig>,

    // --- Proxy-specific flags ---
    pub pinned_model: bool,               // true for `gglib serve`, false for `gglib proxy`
}
```

### `resolve_unified_config()` function

**File:** `crates/gglib-runtime/src/server_config.rs` (new function)

```rust
/// Resolve a [`UnifiedServerConfig`] into a fully-resolved [`ServerConfig`]
/// using a strict 3-tier cascade.
///
/// Tier 1 (Explicit Overrides) > Tier 2 (Model Defaults) > Tier 3 (Global Defaults)
pub fn resolve_unified_config(input: &UnifiedServerConfig) -> ServerConfig {
    // Tier 3 defaults are the floor
    let mut config = ServerConfig::new(
        input.model_id,
        input.model_name.clone(),
        input.model_path.clone(),
        input.global_base_port,
    );

    // --- Context size (4-level cascade) ---
    let ctx = resolve_context_size(&ServerConfigOptions {
        context_size: input.explicit_context_size,
        model_server_ctx: input.model_server_ctx,
        global_default_ctx: input.global_default_ctx,
        ..Default::default()
    });
    config = config.with_context_size(ctx);

    // --- Port ---
    if let Some(port) = input.explicit_port {
        config = config.with_port(port);
    }

    // --- GPU layers (Tier 1 only, no model default yet) ---
    if let Some(layers) = input.explicit_gpu_layers {
        config = config.with_gpu_layers(layers);
    }

    // --- Jinja (Tier 1 > Tier 2 tag-based) ---
    let jinja = resolve_jinja_flag(input.explicit_jinja, &input.model_tags);
    if jinja.enabled {
        config = config.with_jinja();
    }

    // --- Reasoning format (Tier 1 > Tier 2 tag-based) ---
    match input.explicit_reasoning_format.as_deref() {
        Some("none") => { /* suppressed */ }
        Some(format) => config = config.with_reasoning_format(format.to_owned()),
        None => {
            let reasoning = resolve_reasoning_format(None, &input.model_tags);
            if let Some(format) = reasoning.format {
                config = config.with_reasoning_format(format);
            }
        }
    }

    // --- Inference params (Tier 1 explicit > Tier 3 global override) ---
    let inference = input.explicit_inference_params
        .as_ref()
        .or(input.global_inference_override.as_ref())
        .cloned();
    if let Some(params) = inference {
        config = config.with_inference_config(params);
    }

    // --- MTP (Tier 1 > Tier 2 tag-based) ---
    let mtp = resolve_mtp_args(
        input.explicit_mtp_draft_n_max,
        input.explicit_mtp_draft_p_min,
        &input.model_tags,
    );
    if mtp.enabled {
        config = config
            .with_spec_draft_n_max(mtp.draft_n_max)
            .with_spec_draft_p_min(mtp.draft_p_min);
    }

    // --- Memory lock (Tier 1 > Tier 3 default false) ---
    let mlock = input.explicit_mlock.unwrap_or(false);
    if mlock {
        config = config.with_mlock();
    }

    // --- Cache settings (Tier 1 > Tier 3 global) ---
    let slot_save_path = input.explicit_slot_save_path
        .or(input.global_slot_dir);
    config = config.with_slot_save_path(slot_save_path);

    if let Some(mb) = input.explicit_cache_ram_mb {
        config = config.with_cache_ram_mb(mb);
    }

    if let Some(n) = input.explicit_cache_reuse {
        config = config.with_cache_reuse(n);
    }

    let kv_types = resolve_kv_cache_types(
        input.explicit_cache_type_k,
        input.explicit_cache_type_v,
    );
    config = config
        .with_cache_type_k(kv_types.k)
        .with_cache_type_v(kv_types.v);

    config
}
```

### Deprecate `build_server_config()`

Add `#[deprecated(since = "0.x.x", note = "Use resolve_unified_config() instead")]` to `crates/gglib-runtime/src/server_config.rs`. Add `#[allow(deprecated)]` at each remaining call site (`ServerOps`, `agent_chat`, `council`, `plan`, `ProcessManager`) — these will be migrated in Phase 3.

### Tests

- Add a **parity test**: construct both `ServerConfigOptions` and `UnifiedServerConfig` with equivalent inputs, assert that `resolve_unified_config()` produces an identical `ServerConfig` as `build_server_config()`.
- Add a **cascade test**: verify that Tier 1 overrides Tier 2 overrides Tier 3 for context size, jinja, reasoning format, MTP, and mlock.

---

### Problem

`ProcessManager` has two strategies:
- `Concurrent` — multiple simultaneous llama-server processes
- `SingleSwap` — one process, auto-swaps on model change

Neither enforces **single-model pinning** — both accept foreign model requests. For `gglib serve <model>`, we need a mode that starts one model and rejects all others.

### Changes

**File:** `crates/gglib-runtime/src/process/manager.rs`

Add new variant to `ProcessStrategy` (line ~56):

```rust
pub enum ProcessStrategy {
    Concurrent { max_concurrent: usize },
    SingleSwap { /* ... existing fields ... */ },
    /// Single-model pinned mode. Starts one model and rejects all foreign
    /// model requests. Used by `gglib serve <model>`.
    Pinned {
        model_id: i64,
        model_name: String,
        base_port: u16,
        llama_server_path: String,
    },
}
```

Add constructor:

```rust
pub fn new_pinned(
    model_id: i64,
    model_name: String,
    base_port: u16,
    llama_server_path: impl Into<String>,
) -> Self
```

In `ensure_model_running_with()`, add a branch for `Pinned`:

```rust
ProcessStrategy::Pinned { model_name: pinned_name, .. } => {
    if model_name != pinned_name.as_str() {
        return Err(ModelRuntimeError::PinnedModelMismatch {
            expected: pinned_name.clone(),
            requested: model_name.to_owned(),
        });
    }
    // Reuse the SingleSwap startup logic for the pinned model
    // (or inline a simplified version that skips swap logic)
}
```

Add new error variant to `ModelRuntimeError`:

```rust
/// Requested model doesn't match the pinned model in `gglib serve` mode.
PinnedModelMismatch { expected: String, requested: String },
```

**File:** `crates/gglib-core/src/ports/model_runtime.rs` — add the error variant to the core error type.

### Tests

- Test that `ensure_model_running` with the pinned model succeeds
- Test that `ensure_model_running` with a different model returns `PinnedModelMismatch`

---

### Problem

Both GUI bootstraps create two separate process management backends:

| Bootstrap | `ServerOps` uses | `ProxyOps` uses |
|---|---|---|
| Axum | `ProcessRunner` from `CoreBootstrap` (multi-process) | `ModelRuntimePort` wrapping `ProcessManager::SingleSwap` |
| Tauri | `ProcessRunner` from `CoreBootstrap` (multi-process) | `ModelRuntimePort` wrapping `ProcessManager::SingleSwap` |

For unification, the GUI must **always** go through the proxy pipeline.

#### 3a. Add `ProxyOps` dependency to `ServerDeps`

**File:** `crates/gglib-app-services/src/servers.rs`

```rust
pub struct ServerDeps {
    pub core: Arc<AppCore>,
    pub runner: Arc<dyn ProcessRunner>,          // keep for backward compat
    pub proxy: Arc<ProxyOps>,                    // NEW: unified path
    pub emitter: Arc<dyn AppEventEmitter>,
    pub server_events: Arc<dyn gglib_core::events::ServerEvents>,
    pub tool_detector: Arc<dyn ToolSupportDetectorPort>,
}
```

#### 3b. Modify `ServerOps::start()` to delegate to proxy

In `ServerOps::start()`, replace the direct `runner.start(config)` path with:

```rust
pub async fn start(&self, id: i64, request: StartServerRequest) -> Result<StartServerResponse, GuiError> {
    // 1. Resolve model and build UnifiedServerConfig
    let model = self.deps.core.models().get_by_id(id).await?;
    let unified = UnifiedServerConfig::from_start_request(&request, &model, &settings).await?;
    let server_config = resolve_unified_config(&unified);

    // 2. Ensure proxy is running (idempotent)
    let proxy_addr = self.deps.proxy.ensure_running().await?;

    // 3. Ensure model is loaded in the proxy's manager
    let target = self.deps.proxy.runtime().ensure_model_running(
        &model.name,
        server_config.context_size,
        unified.global_default_ctx.unwrap_or(DEFAULT_CONTEXT_SIZE),
    ).await?;

    Ok(StartServerResponse {
        address: proxy_addr,
        model_id: id,
        // ...
    })
}
```

#### 3c. Add `ensure_running()` and `runtime()` to `ProxyOps`

**File:** `crates/gglib-app-services/src/proxy.rs`

```rust
/// Start the proxy if not already running, or return the existing address.
pub async fn ensure_running(&self) -> Result<SocketAddr, GuiError> {
    if let Some(status) = self.supervisor.status() {
        if status.is_running() {
            return Ok(status.address);
        }
    }
    self.start(ProxyConfig::default()).await
}

/// Access the underlying model runtime.
pub fn runtime(&self) -> &Arc<dyn ModelRuntimePort> {
    &self.runtime
}
```

#### 3d. Update Axum Bootstrap

**File:** `crates/gglib-axum/src/bootstrap.rs`

Reorder construction so `ProxyOps` is created before `ServerOps`, then pass it:

```rust
// Create proxy BEFORE servers (reorder from current)
let proxy = Arc::new(ProxyOps::new(ProxyDeps { /* ... */ }));

let servers = Arc::new(ServerOps::new(ServerDeps {
    core: Arc::clone(&core),
    runner: runner.clone(),
    proxy: proxy.clone(),    // NEW
    emitter: sse_emitter,
    server_events,
    tool_detector: tool_detector.clone(),
}));
```

#### 3e. Update Tauri Bootstrap

**File:** `crates/gglib-tauri/src/bootstrap.rs`

Same reordering as Axum, at the equivalent location.

---

### Problem

The current `gglib serve` handler (`crates/gglib-cli/src/handlers/inference/serve.rs`) uses `LlamaCommandBuilder` to directly spawn `llama-server`, bypassing the proxy entirely. This means no dashboard, no caching, no SSE, no `--host 127.0.0.1`, no `--metrics`, no `--parallel 1`.

### Changes

**File:** `crates/gglib-cli/src/handlers/inference/serve.rs`

Replace the entire `execute()` body with:

```rust
pub async fn execute(
    ctx: &CliContext,
    id: u32,
    context: ContextArgs,
    options: ServeOptions,
    sampling: SamplingArgs,
    mtp: MtpArgs,
    verbose: bool,
) -> Result<()> {
    // 1. Resolve model
    let model = ctx.app.models().get_by_id(id as i64).await?;

    // 2. Build UnifiedServerConfig from CLI args + model metadata + global defaults
    let unified = UnifiedServerConfig {
        model_id: model.id,
        model_name: model.name.clone(),
        model_path: model.path.clone(),
        model_tags: model.tags.clone(),
        explicit_context_size: context.context_size,
        explicit_mlock: context.mlock.then_some(true),
        explicit_inference_params: resolve_inference_config(&sampling).await,
        explicit_mtp_draft_n_max: mtp.draft_n_max,
        explicit_mtp_draft_p_min: mtp.draft_p_min,
        // ... all other fields from CLI args ...
        pinned_model: true,  // KEY: this is serve, not proxy
        // ... global defaults from settings ...
    };

    // 3. Resolve config through 3-tier cascade
    let server_config = resolve_unified_config(&unified);

    // 4. Create Pinned ProcessManager
    let llama_path = resolve_llama_server(&ctx.bootstrap).await?;
    let manager = ProcessManager::new_pinned(
        model.id,
        model.name.clone(),
        DEFAULT_LLAMA_BASE_PORT,
        llama_path,
    );
    let runtime = Arc::new(RuntimePortImpl::new(manager));

    // 5. Start the proxy with the pinned runtime
    let config = ProxyConfig {
        host: "127.0.0.1".to_string(),
        port: options.port.unwrap_or(DEFAULT_PROXY_PORT),
        ..Default::default()
    };

    // 6. Initialize proxy supervisor and start
    let supervisor = ProxySupervisor::new();
    let addr = supervisor.start(config, runtime, /* ... catalog, mcp, council ... */).await?;

    println!("Serving {} on http://{}", model.name, addr);

    // 7. Wait for shutdown signal
    tokio::signal::ctrl_c().await?;
    println!("Shutting down...");
    Ok(())
}
```

### Key behavioral changes (intentional, per Rule 1: No backwards compatibility)

| Behavior | Before (`LlamaCommandBuilder`) | After (unified proxy) |
|---|---|---|
| Host binding | ❌ Missing (security risk) | ✅ `127.0.0.1` |
| Metrics | ❌ Missing | ✅ `--metrics` |
| Parallel limit | ❌ Missing | ✅ `--parallel 1` |
| KV cache | ❌ Missing | ✅ Full cache lifecycle |
| Dashboard | ❌ Missing | ✅ `/v1/proxy/status` |
| SSE streaming | ❌ Missing | ✅ Prompt ingest progress |
| Startup latency | ~0s | ~1-3s (acceptable) |
| Memory overhead | ~0MB | ~20-50MB (proxy runtime) |

---

### Changes

**File to delete:** `crates/gglib-runtime/src/llama/invocation.rs`

After `serve.rs` is rewritten (Step 4), `LlamaCommandBuilder` has zero remaining callers. Delete the entire file and remove the module declaration from `crates/gglib-runtime/src/llama/mod.rs`.

Verify with:
```bash
grep -r "LlamaCommandBuilder" crates/ --include="*.rs"
# Should return zero results
```

---

## 6. Verification Gates

After completing all changes:

- [ ] `cargo check --workspace` passes
- [ ] `cargo clippy --workspace -- -D warnings` passes (with `#[allow(deprecated)]` at remaining `build_server_config()` call sites)
- [ ] `cargo test -p gglib-core` passes (new `UnifiedServerConfig` module)
- [ ] `cargo test -p gglib-runtime` passes (new `resolve_unified_config`, `Pinned` strategy, parity tests)
- [ ] `cargo test -p gglib-app-services` passes (ServerOps delegation to ProxyOps)
- [ ] `cargo test -p gglib-axum` passes (bootstrap reordering)
- [ ] `cargo test -p gglib-tauri` passes (bootstrap reordering)
- [ ] `cargo test -p gglib-cli` passes (serve.rs rewrite)
- [ ] Verify `LlamaCommandBuilder` has zero remaining references
- [ ] Manual test: `gglib serve <model>` starts proxy with dashboard accessible at `/v1/proxy/status`
- [ ] Manual test: `gglib serve <model>` rejects requests for different models with `PinnedModelMismatch`

---

## 7. Files Summary

| Action | File | Change |
|---|---|---|
| **Create** | `crates/gglib-core/src/unified_server_config.rs` | `UnifiedServerConfig` struct + module |
| **Modify** | `crates/gglib-core/src/lib.rs` | Add `unified_server_config` module |
| **Modify** | `crates/gglib-runtime/src/server_config.rs` | Add `resolve_unified_config()`, deprecate `build_server_config()` |
| **Modify** | `crates/gglib-runtime/src/process/manager.rs` | Add `ProcessStrategy::Pinned`, `new_pinned()`, `PinnedModelMismatch` error |
| **Modify** | `crates/gglib-core/src/ports/model_runtime.rs` | Add `PinnedModelMismatch` to error type |
| **Modify** | `crates/gglib-app-services/src/servers.rs` | Add `ProxyOps` to `ServerDeps`, delegate `start()` to proxy |
| **Modify** | `crates/gglib-app-services/src/proxy.rs` | Add `ensure_running()` and `runtime()` methods |
| **Modify** | `crates/gglib-axum/src/bootstrap.rs` | Reorder: create `ProxyOps` before `ServerOps`, wire `proxy` into `ServerDeps` |
| **Modify** | `crates/gglib-tauri/src/bootstrap.rs` | Same as Axum |
| **Rewrite** | `crates/gglib-cli/src/handlers/inference/serve.rs` | Replace `LlamaCommandBuilder` with unified proxy stack |
| **Delete** | `crates/gglib-runtime/src/llama/invocation.rs` | Remove `LlamaCommandBuilder` (zero callers after serve rewrite) |
| **Modify** | `crates/gglib-runtime/src/llama/mod.rs` | Remove `invocation` module declaration |

---

## 8. Out of Scope (Phase 3)

- Deleting `ServerOps` entirely (Phase 3 — keep for backward compat in Phase 2)
- Adding KV cache persistence flags to CLI `serve` (Phase 3)
- Unifying `SamplingArgs` usage across CLI and GUI (Phase 3)
- Updating CLI help strings for `gglib serve` (Phase 3)
- CI integration tests for `/v1/proxy/status` on `gglib serve` (Phase 3)

### LINKED ISSUE #633: Phase 3: Documentation, CLI Parity & CI Alignment
# Phase 3: CLI Parity, CI & Docs

**Parent Epic:** [#630](https://github.com/mmogr/gglib/issues/630) — Unify `gglib serve` and `gglib proxy` into a Single Execution Path
**Depends on:** [#631](https://github.com/mmogr/gglib/issues/631) (Phase 1: Foundation), [#632](https://github.com/mmogr/gglib/issues/632) (Phase 2: Pinned Mode & Serve Rewrite)

## Objective

This is the final cleanup phase. After Phase 2 rewrote `gglib serve` to use the proxy stack, Phase 3 ensures:

1. **CLI parity** — `gglib serve` exposes the same cache flags (`--cache`, `--cache-ram-mb`, `--cache-reuse`, etc.) that `gglib proxy` already has.
2. **Shared arg unification** — `SamplingArgs` and cache args are extracted into reusable shared structs so both commands parse identically.
3. **Help text accuracy** — `gglib serve --help` explicitly mentions the dashboard and KV caching.
4. **CI integration tests** — automated tests verify the dashboard endpoint and KV cache persistence on the serve path.
5. **Documentation updates** — `README.md` reflects that `serve` now has full proxy parity.

---

### Problem

Currently all 7 cache-related CLI flags live **only** on the `Proxy` command as inline fields (`commands.rs`). The `Serve` command has **zero** cache flags. These args are not extracted into a reusable shared struct, so `Serve` can't inherit them via `#[command(flatten)]`.

#### 1a. Create `CacheArgs` shared struct

**File:** `shared_args.rs`

Add a new shared args struct after `MtpArgs` (after line ~88):

```rust
/// Shared cache-related CLI arguments.
///
/// Flattened into both `Serve` and `Proxy` commands so cache tuning is
/// available regardless of which entry point the user chooses.
#[derive(Args, Debug, Clone, Default)]
pub struct CacheArgs {
    /// Enable KV cache session persistence (disk slot save/restore)
    #[arg(long)]
    pub cache: bool,

    /// Directory for KV cache slot files (default: <data-root>/slots)
    #[arg(long)]
    pub slot_dir: Option<std::path::PathBuf>,

    /// RAM budget in MiB for llama-server's host-RAM prompt cache
    /// (default: auto-sized from system RAM and model weights)
    #[arg(long)]
    pub cache_ram_mb: Option<u64>,

    /// Minimum chunk size in tokens for KV-shift cache reuse
    /// (default: 0, disabled)
    #[arg(long)]
    pub cache_reuse: Option<u32>,

    /// Disk budget in GiB for KV cache slot file eviction
    /// (default: auto-derived from free disk space)
    #[arg(long)]
    pub cache_disk_gb: Option<u64>,

    /// K cache element type (default: q8_0)
    #[arg(long, value_enum)]
    pub cache_type_k: Option<gglib_core::cache_config::KvCacheType>,

    /// V cache element type (default: q8_0)
    #[arg(long, value_enum)]
    pub cache_type_v: Option<gglib_core::cache_config::KvCacheType>,
}
```

Add the required import at the top of the file:
```rust
use gglib_core::cache_config::KvCacheType;
```

#### 1b. Flatten `CacheArgs` into `Serve` command

**File:** `commands.rs`

Update the `Serve` variant (lines 155–165) to include `CacheArgs`:

```rust
/// Serve a GGUF model with a pinned single-model proxy server.
///
/// Starts a full proxy stack (caching, dashboard, request normalization)
/// with the specified model pinned — it cannot be swapped at runtime.
/// Use `gglib proxy` for dynamic model selection.
#[command(display_order = 10)]
Serve {
    /// ID of the model to serve
    id: u32,
    #[command(flatten)]
    context: ContextArgs,
    #[command(flatten)]
    options: ServeOptions,
    #[command(flatten)]
    sampling: SamplingArgs,
    #[command(flatten)]
    mtp: MtpArgs,
    #[command(flatten)]
    cache: CacheArgs,  // NEW
},
```

#### 1c. Replace inline cache fields on `Proxy` with `CacheArgs`

**File:** `commands.rs`

Replace the 7 inline cache fields on the `Proxy` variant (lines 494–547) with a single flattened struct:

```rust
// REMOVE these inline fields:
//   cache: bool,
//   slot_dir: Option<std::path::PathBuf>,
//   cache_ram_mb: Option<u64>,
//   cache_reuse: Option<u32>,
//   cache_disk_gb: Option<u64>,
//   cache_type_k: Option<KvCacheType>,
//   cache_type_v: Option<KvCacheType>,

// REPLACE with:
#[command(flatten)]
cache_args: CacheArgs,
```

Update the `Proxy` dispatch in `dispatch.rs` to reference `command.cache_args.cache` instead of `command.cache`, etc.

---

### Problem

The `Proxy` command defines sampling parameters as **inline fields** (`commands.rs`):

```rust
#[arg(long)] temperature: Option<f32>,
#[arg(long)] top_p: Option<f32>,
// ...
```

Meanwhile `Serve` uses the shared `shared_args.rs` struct with the **same fields** but slightly different clap attribute names. This creates a subtle inconsistency.

#### 2a. Replace inline sampling fields on `Proxy` with `SamplingArgs`

**File:** `commands.rs`

Remove the 7 inline sampling fields from `Proxy` (lines 476–491) and replace with:

```rust
#[command(flatten)]
sampling: SamplingArgs,
```

Update the `Proxy` dispatch in `dispatch.rs` to construct `InferenceConfig` from `command.sampling` instead of the individual fields.

#### 2b. Verify `SamplingArgs` flag names are consistent

**File:** `shared_args.rs`

The existing `SamplingArgs` uses `long = "top-p"` style names (kebab-case). Verify these match what users expect. The result should be identical — confirm by running `gglib proxy --help` and `gglib serve --help` and comparing the sampling section.

---

### Problem

The current `Serve` command help text (`commands.rs`) is misleading after the Phase 2 rewrite since it no longer directly invokes llama-server.

#### 3a. Update `Serve` command doc comment

**File:** `commands.rs`

Replace the doc comment at line 155:

```rust
/// Serve a GGUF model with a pinned single-model proxy server.
///
/// This starts a full proxy stack (caching, dashboard, request normalization)
/// with the specified model pinned — it cannot be swapped at runtime.
///
/// Dashboard:  [http://127.0.0.1](http://127.0.0.1):<port>/v1/proxy/status
/// API:        [http://127.0.0.1](http://127.0.0.1):<port>/v1/chat/completions
/// Models:     [http://127.0.0.1](http://127.0.0.1):<port>/v1/models
///
/// Use `gglib proxy` for dynamic multi-model selection.
```

#### 3b. Update `ServeOptions` help text

**File:** `shared_args.rs`

Add help text to the `port` field in `ServeOptions` (line ~94):

```rust
/// Port for the proxy server (default: 8080).
/// Dashboard available at [http://127.0.0.1](http://127.0.0.1):<port>/v1/proxy/status
#[arg(short, long, default_value = "8080")]
pub port: u16,
```

---

### Problem

There are **no integration tests** for `gglib serve` or `gglib proxy` functionality. The CI runs `cargo test -p gglib-proxy` (`ci.yml`) which exercises only unit tests inside the crate.

#### 4a. Create `tests/integration_serve_proxy.rs`

**File to create:** `tests/integration_serve_proxy.rs`

```rust
//! Integration tests for the unified serve/proxy stack.
//!
//! These tests verify that `gglib serve` (pinned mode) exposes the same
//! endpoints and caching behavior as `gglib proxy`.

mod integration_serve_proxy {

    use std::time::Duration;
    use reqwest::Client;
    use serde_json::Value;

    const PROXY_STARTUP_TIMEOUT: Duration = Duration::from_secs(30);
    const HEALTH_POLL_INTERVAL: Duration = Duration::from_secs(1);

    /// Wait for the proxy dashboard endpoint to become available.
    async fn wait_for_dashboard(base_url: &str) -> anyhow::Result<()> {
        let client = Client::new();
        let deadline = std::time::Instant::now() + PROXY_STARTUP_TIMEOUT;

        while std::time::Instant::now() < deadline {
            match client.get(format!("{}/v1/proxy/status", base_url)).send().await {
                Ok(resp) if resp.status().is_success() => return Ok(()),
                Ok(resp) => eprintln!("Dashboard returned {}: {}", resp.status(), resp.status()),
                Err(e) if e.is_connect() => { /* not ready yet */ }
                Err(e) => return Err(e.into()),
            }
            tokio::time::sleep(HEALTH_POLL_INTERVAL).await;
        }
        Err(anyhow::anyhow!("Dashboard did not become available within {:?}", PROXY_STARTUP_TIMEOUT))
    }

    /// Test: `gglib serve` exposes the dashboard endpoint.
    #[tokio::test]
    #[ignore] // Requires a model file; run with: cargo test integration_serve_proxy -- --ignored
    async fn test_serve_dashboard_is_alive() {
        // 1. Start the proxy in pinned mode (using the same code path as `gglib serve`)
        let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
        let handle = tokio::spawn(async move {
            // Build UnifiedServerConfig for a test model
            // Create ProcessManager::new_pinned()
            // Start ProxySupervisor
            // Wait for shutdown signal
        });

        let base_url = "[http://127.0.0.1:18080](http://127.0.0.1:18080)";

        // 2. Wait for dashboard
        wait_for_dashboard(base_url).await.expect("Dashboard should become available");

        // 3. Verify response structure
        let client = Client::new();
        let resp: Value = client
            .get(format!("{}/v1/proxy/status", base_url))
            .send()
            .await
            .expect("Dashboard request should succeed")
            .json()
            .await
            .expect("Dashboard response should be valid JSON");

        // Verify expected top-level fields from DashboardSnapshot
        assert!(resp.get("active_connections").is_some(), "Missing active_connections");
        assert!(resp.get("slots").is_some(), "Missing slots");
        assert!(resp.get("total_requests").is_some(), "Missing total_requests");
        assert!(resp.get("upstream_health").is_some(), "Missing upstream_health");

        // 4. Shutdown
        let _ = shutdown_tx.send(true);
        handle.await.unwrap();
    }

    /// Test: KV cache persistence works on the serve endpoint.
    #[tokio::test]
    #[ignore] // Requires a model file
    async fn test_serve_kv_cache_persistence() {
        let slot_dir = std::env::temp_dir().join(format!("gglib-test-slots-{}", std::process::id()));
        std::fs::create_dir_all(&slot_dir).unwrap();

        // 1. Start proxy with cache enabled
        let (shutdown_tx, _) = tokio::sync::watch::channel(false);
        // ... start proxy with cache_enabled=true, slot_dir=Some(slot_dir) ...

        let base_url = "[http://127.0.0.1:18081](http://127.0.0.1:18081)";
        wait_for_dashboard(base_url).await.unwrap();

        let client = Client::new();

        // 2. Send first request (cache miss)
        let first_resp: Value = client
            .post(format!("{}/v1/chat/completions", base_url))
            .json(&serde_json::json!({
                "model": "test-model",
                "messages": [{"role": "user", "content": "Hello, this is a cache test."}],
                "session_id": "cache-test-session"
            }))
            .send()
            .await
            .unwrap()
            .json()
            .await
            .unwrap();

        let cache_hits_1 = first_resp
            .get("usage")
            .and_then(|u| u.get("prompt_tokens_details"))
            .and_then(|d| d.get("prompt_cache_hits"))
            .and_then(|h| h.as_u64())
            .unwrap_or(0);
        assert_eq!(cache_hits_1, 0, "First request should be a cache miss");

        // 3. Wait for slot file to be written
        tokio::time::sleep(Duration::from_secs(2)).await;

        let slot_files = std::fs::read_dir(&slot_dir)
            .unwrap()
            .filter_map(|e| e.ok())
            .filter(|e| e.path().extension().map(|ext| ext == "bin").unwrap_or(false))
            .count();
        assert!(slot_files > 0, "Slot file should have been written to disk");

        // 4. Send second request with same session (should be cache hit)
        let second_resp: Value = client
            .post(format!("{}/v1/chat/completions", base_url))
            .json(&serde_json::json!({
                "model": "test-model",
                "messages": [
                    {"role": "user", "content": "Hello, this is a cache test."},
                    {"role": "assistant", "content": "First response."},
                    {"role": "user", "content": "Follow-up question."}
                ],
                "session_id": "cache-test-session"
            }))
            .send()
            .await
            .unwrap()
            .json()
            .await
            .unwrap();

        let cache_hits_2 = second_resp
            .get("usage")
            .and_then(|u| u.get("prompt_tokens_details"))
            .and_then(|d| d.get("prompt_cache_hits"))
            .and_then(|h| h.as_u64())
            .unwrap_or(0);
        assert!(cache_hits_2 > 0, "Second request should have cache hits, got {}", cache_hits_2);

        // 5. Cleanup
        let _ = shutdown_tx.send(true);
        let _ = std::fs::remove_dir_all(&slot_dir);
    }
}
```

#### 4b. Add test job to CI workflow

**File:** `ci.yml`

Add a new job after the existing `test` job:

```yaml
test-integration-serve:
  runs-on: ubuntu-latest
  timeout-minutes: 15
  steps:
    - uses: actions/checkout@v4
    - uses: dtolnay/rust-toolchain@stable
    - uses: Swatinem/rust-cache@v2
    - name: Run serve/proxy integration tests
      run: |
        # These tests are marked #[ignore] and require a model file.
        # We run them against a small test model if available, or skip gracefully.
        cargo test -p gglib --test integration_serve_proxy -- --ignored --test-threads=1 2>&1 || \
          echo "::warning::Integration serve tests skipped (no test model available)"
```

---

#### 5a. Update the `gglib serve` example

Update the quick-start example (around line 28) to mention the dashboard:

```markdown
gglib serve qwen2.5

# Dashboard: [http://127.0.0.1:8080/v1/proxy/status](http://127.0.0.1:8080/v1/proxy/status)
# API:      [http://127.0.0.1:8080/v1/chat/completions](http://127.0.0.1:8080/v1/chat/completions)
```

#### 5b. Update the caching section

Update the caching section header (line 203) to:

```markdown
### KV Cache Session Persistence

Available on both `gglib serve` and `gglib proxy`. Enable with `--cache`:
```
*(Note that `gglib serve` enables caching by default.)*

#### 5c. Update the dashboard section

Update the dashboard section (line 190) to:

```markdown
**Dashboard** — available on both `gglib serve` and `gglib proxy`. A unified `DashboardSnapshot` of active connections, per-slot context usage, KV cache hits, and recent request history.
```

#### 5d. Remove any "proxy-only" caveats

Search the `README.md` for phrases like "proxy only", "when using proxy", or "gglib proxy provides" and update them to reflect that these features are now available on `serve` as well.

---

## 6. Verification Gates

After completing all changes:

- [ ] `cargo check --workspace` passes
- [ ] `cargo clippy --workspace -- -D warnings` passes
- [ ] `cargo test --workspace` passes (including new `integration_serve_proxy` tests, which are `#[ignore]` by default)
- [ ] `gglib serve --help` shows cache flags (`--cache`, `--cache-ram-mb`, `--cache-reuse`, `--cache-disk-gb`, `--cache-type-k`, `--cache-type-v`)
- [ ] `gglib serve --help` mentions the dashboard URL and KV caching in the command description
- [ ] `gglib proxy --help` still shows all the same flags (no regression from extracting `CacheArgs` and `SamplingArgs`)
- [ ] `gglib serve --help` and `gglib proxy --help` show identical sampling flags (`--temperature`, `--top-p`, `--top-k`, etc.)
- [ ] `README.md` no longer describes caching or dashboard as proxy-only features

---

## 7. Files Summary

| Action | File | Lines |
|---|---|---|
| **Modify** | `shared_args.rs` | Add `CacheArgs` struct (after `MtpArgs` ~line 88), update `ServeOptions` help text |
| **Modify** | `commands.rs` | Add `CacheArgs` flatten to `Serve` (~line 165), replace inline cache/sampling fields on `Proxy` with flattened structs, update `Serve` doc comment |
| **Modify** | `dispatch.rs` | Update `Proxy` dispatch to use `command.cache_args` and `command.sampling` instead of inline fields |
| **Create** | `tests/integration_serve_proxy.rs` | New integration test file (dashboard alive test, KV cache persistence test) |
| **Modify** | `ci.yml` | Add `test-integration-serve` job |
| **Modify** | `README.md` | Update serve example, caching section, dashboard section, remove proxy-only caveats |

mmogr:
### Before — two divergent paths

```mermaid
flowchart TD
    subgraph BEFORE["Before #630: gglib serve bypasses the proxy entirely"]
        direction TB

        S1["gglib serve &lt;model&gt;"] --> INV["invocation.rs<br/>(LlamaCommandBuilder)"]
        INV --> RAW["cmd.status()<br/>raw process spawn"]
        RAW --> LS1["llama-server<br/>(direct, inherited stdio)"]

        P1["gglib proxy"] --> SC["server_config.rs"]
        GA1["Axum GUI"] --> SC
        GT1["Tauri GUI"] --> SC
        SC --> CMD["command.rs<br/>(canonical flag builder)"]
        CMD --> MGR["manager.rs<br/>ProcessManager<br/>(auto-start / swap / health)"]
        MGR --> AX["Axum HTTP proxy<br/>server.rs"]
        AX --> LS2["llama-server<br/>(managed)"]
        AX --> DASH["Dashboard<br/>/v1/proxy/status + SSE"]
        AX --> CACHE["cache_lifecycle.rs<br/>restore → forward → save"]
    end

    style LS1 fill:#5a1f1f,stroke:#c0392b
    style S1 fill:#5a1f1f,stroke:#c0392b
```

**Gap:** `serve`'s path has no `--host 127.0.0.1` default (security risk), no `--metrics`, no `--parallel 1`, no dashboard, no KV cache — two config structs (`ProxyConfig` vs `ServerConfigOptions`) drifting apart.

---

### After — one unified path

```mermaid
flowchart TD
    subgraph AFTER["After #630: every surface shares one engine"]
        direction TB

        S2["gglib serve &lt;model&gt;"] --> UNI["UnifiedServerConfig<br/>3-tier cascade:<br/>explicit → model defaults → global floor"]
        P2["gglib proxy"] --> UNI
        GA2["Axum GUI (ServerOps)"] --> UNI
        GT2["Tauri GUI (ServerOps)"] --> UNI

        UNI --> BSC["build_server_config()"]
        BSC --> CMD2["command.rs<br/>single flag builder<br/>--host 127.0.0.1 · --metrics · --parallel 1<br/>cache flags · --mlock"]
        CMD2 --> MGR2["ProcessManager (SingleSwap)<br/>new_pinned() rejects foreign models"]
        MGR2 --> AX2["gglib-proxy Axum HTTP layer<br/>server.rs"]
        AX2 --> LS3["llama-server<br/>(one instance, system-wide)"]
        AX2 --> DASH2["Dashboard<br/>/v1/proxy/status + SSE"]
        AX2 --> CACHE2["KV cache lifecycle<br/>restore → forward → save"]
        AX2 --> FILT["/v1/models<br/>filtered to pinned model<br/>(pinned mode only)"]
    end

    style UNI fill:#1f3a5a,stroke:#2e86de
    style CMD2 fill:#1f3a5a,stroke:#2e86de
```

**Result:** `gglib serve` is now a pinned invocation of the same proxy stack — same dashboard, same cache, same hardened defaults, same command builder — that `gglib proxy` and both GUIs already used.

This is the umbrella PR for the serve/proxy parity epic. Individual phases were merged into this branch as PRs; this PR merges the finished branch into `main`.

### Completed Phases

**Phase 1: Foundation (mlock plumbing)** — Merged via #635
- Wired the missing `mlock` field through the canonical config pipeline
- Frontend `mlock` → `StartServerRequest` → `ServerConfigOptions` → `build_server_config()` → `ServerConfig` → `build_command()` → `--mlock` flag

```mermaid
flowchart LR
    FE["Frontend<br/>ServeConfig.mlock"] -->|"JSON"| SSR["StartServerRequest<br/>types.rs"]
    SSR -->|"❌ field didn't exist<br/>serde silently drops it"| GAP["🚫 dropped"]
    SSR -.->|"✅ added"| SCO["ServerConfigOptions<br/>server_config.rs"]
    SCO -->|"✅ added"| SC["ServerConfig<br/>process_runner.rs"]
    SC -->|"✅ added"| BC["build_command()<br/>command.rs"]
    BC -->|"✅ added"| FLAG["--mlock<br/>on the llama-server invocation"]

    style GAP fill:#5a1f1f,stroke:#c0392b
    style SCO fill:#1f3a5a,stroke:#2e86de
    style SC fill:#1f3a5a,stroke:#2e86de
    style BC fill:#1f3a5a,stroke:#2e86de
    style FLAG fill:#1f3a5a,stroke:#2e86de
```

**Achieved:** closed the one gap where a GUI setting reached the frontend request but was silently discarded before it could ever influence a launch. No behavior change for existing callers — purely wiring a previously-dead field through the canonical pipeline.

**Phase 2: Pinned Mode & Serve Rewrite** — Merged via #637 (resolves #632)
- Introduced `UnifiedServerConfig` and the 3-tier config cascade
- `ProcessManager::new_pinned` / `ProcessStrategy::Pinned`, rejecting foreign-model requests
- Rewrote `gglib serve` to run through the unified proxy stack instead of a direct `llama-server` spawn
- Routed `ServerOps` through `ProxyOps` in both the Axum and Tauri bootstraps
- Deleted `LlamaCommandBuilder`

```mermaid
flowchart TD
    subgraph OLD["Removed"]
        INV["invocation.rs<br/>LlamaCommandBuilder"]
        RAW["cmd.status()<br/>raw spawn, no HTTP layer"]
        INV --> RAW
    end

    S["gglib serve &lt;model&gt;"] -.->|"rewritten off<br/>the old path"| OLD
    S --> UNI["🆕 UnifiedServerConfig<br/>3-tier cascade<br/>explicit → model defaults → global floor"]
    UNI --> BSC["build_server_config()"]
    BSC --> PIN["🆕 ProcessManager::new_pinned()<br/>rejects foreign-model requests"]
    PIN --> PROXY["gglib-proxy stack<br/>(Axum, dashboard, cache lifecycle)"]

    GA["Axum bootstrap<br/>ServerOps"] -->|"🆕 routed through"| PO["ProxyOps"]
    GT["Tauri bootstrap<br/>ServerOps"] -->|"🆕 routed through"| PO
    PO --> PIN

    style OLD fill:#5a1f1f,stroke:#c0392b
    style UNI fill:#1f3a5a,stroke:#2e86de
    style PIN fill:#1f3a5a,stroke:#2e86de
    style PO fill:#1f3a5a,stroke:#2e86de
```

**Achieved:** `gglib serve` stopped being a bare-metal `llama-server` spawn and became a pinned invocation of the real proxy stack — gaining the dashboard, KV cache lifecycle, and request normalization it never had. Both GUI adapters (Axum, Tauri) were cut over to drive models through the same `ProxyOps`/`ProcessManager` instead of a separate direct-spawn path. `LlamaCommandBuilder` deleted outright — no shim, no fallback.

**Phase 3: Catalog Filtering, CLI Parity, CI & Docs** — Merged via #638 (resolves #633)
- Filtered `/v1/models` to the pinned model on the `gglib serve` path
- Extracted shared `CacheArgs`/`SamplingArgs` so `serve` and `proxy` parse identically
- CLI parity tests, updated `--help` text, updated READMEs

```mermaid
flowchart TD
    subgraph SHARED["🆕 shared_args.rs"]
        CA["CacheArgs"]
        SA["SamplingArgs"]
    end

    CA -->|"#[command(flatten)]"| SERVE["Serve command"]
    CA -->|"#[command(flatten)]"| PROXY["Proxy command"]
    SA -->|"#[command(flatten)]"| SERVE
    SA -->|"replaces 7 inline dupes"| PROXY

    PIN["Pinned proxy runtime"] --> MODELS["/v1/models"]
    MODELS -->|"🆕 filtered to"| ONE["just the pinned model<br/>+ council virtuals"]

    HELP["🆕 --help text"] --> SERVE
    HELP -.->|"documents dashboard + caching"| PROXY

    CI["🆕 CI coverage"] --> DASH_T["dashboard reachable on serve"]
    CI --> CACHE_T["KV cache persists on the pinned path"]

    style SHARED fill:#1f3a5a,stroke:#2e86de
    style ONE fill:#1f3a5a,stroke:#2e86de
    style CI fill:#1f3a5a,stroke:#2e86de
```

**Achieved:** `serve` and `proxy` now parse cache and sampling flags from the exact same structs — verified byte-identical in `--help` output. BYOK clients hitting `/v1/models` on a pinned endpoint see only the pinned model, not the full catalog. Help text stopped describing the old direct-spawn behavior. New integration tests assert the dashboard and cache persistence actually work over real HTTP on the pinned path, not just at the config-object level.

**Audit cleanup** — Merged via #639
- A post-merge audit of phases 1–3 found 11 issues, including one real regression (`ModelOps` reading a process registry `ServerOps` had already moved off of, breaking the force-delete guard for a running model)
- Also: deleted `resolve_unified_config` and `ProcessStrategy::Concurrent` (both dead after the phase 2 cutover), closed a GUI proxy-start port collision, wired KV cache flags through the GUI proxy-start endpoint, added `--host` to `gglib serve`, gave `gglib proxy` its own CLI handler module, added the dashboard URL/cache state to the startup banner, and added HTTP-level test coverage for pin enforcement and KV cache persistence on the pinned path

### Verification

- `cargo fmt -- --check` ✅
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` ✅
- `cargo test --workspace` ✅ (1993 passed, 0 failed as of the last audit-cleanup commit)

Closes #630


crates/gglib-app-services/src/service_graph.rs
```diff
@@ -0,0 +1,267 @@
...
+        approval_registry,
+        council_repo,
+        runtime: Arc::clone(&runtime),
+    }));
+
+    let models = Arc::new(ModelOps::new(ModelDeps {
+        core: Arc::clone(&core),
+        runner: Arc::clone(&runner),
+        gguf_parser,
+    }));
```
mmogr, line 221: This is where the divergence gets wired in, so flagging it here rather than in `models.rs` (untouched by this PR).

`ServerOps` (line 216) now starts models through `proxy` → `ProxyOps::runtime()` → `ProcessManager`, which owns its own `GuiProcessCore` map. `ModelOps` (line 210) still receives the bootstrap's `runner`, a `LlamaServerRunner` with a **separate** `ProcessCore` map. On `main` both received the same `runner.clone()`, so `ModelOps::get_server_status` worked; after this PR the two registries never see each other's processes.

Two consequences:

1. `models.rs:38` — `is_serving`/`port` are now permanently `false` on all six DTO paths (`list`, `list_with_query`, `get`, `get_detail`, …). Latent today: `src/types/index.ts:87` declares `isServing` but nothing in the frontend reads it. It will bite whoever wires it up.
2. `models.rs:176` — **live bug.** The `remove` force guard calls `find_handle(&*self.deps.runner, id)`, which now always returns `None`. Deleting a model that is running under the proxy no longer returns `Conflict`, no longer stops the server, and leaves an orphaned `llama-server` serving a catalog entry that no longer exists.

`ModelRuntimePort::list_running()` already returns `Vec<ProcessHandle>` (added in 492106d for exactly this reason), so the fix is a dependency swap rather than new plumbing.

### Target state (not directly committable — spans 3 files)

```rust
// crates/gglib-app-services/src/models.rs — ModelDeps
pub struct ModelDeps {
    pub core: Arc<AppCore>,
    /// The shared proxy runtime, not a standalone runner: `ServerOps` starts
    /// models through the `ProcessManager`, so that is the only registry that
    /// knows what is actually running.
    pub runtime: Arc<dyn ModelRuntimePort>,
    pub gguf_parser: Arc<dyn GgufParserPort>,
}

// models.rs:37 — get_server_status
async fn get_server_status(&self, model_id: i64) -> (bool, Option<u16>) {
    self.deps
        .runtime
        .list_running()
        .await
        .iter()
        .find(|h| h.model_id == model_id)
        .map_or((false, None), |h| (true, Some(h.port)))
}

// models.rs:176 — remove guard, via the same runtime
if let Some(handle) = self
    .deps
    .runtime
    .list_running()
    .await
    .into_iter()
    .find(|h| h.model_id == id)
{
    if !request.force {
        return Err(GuiError::Conflict(format!(
            "Model is currently serving on port {}. Stop the server first or use force=true",
            handle.port
        )));
    }
    self.deps
        .runtime
        .stop_current()
        .await
        .map_err(|e| GuiError::Internal(format!("Failed to stop server: {e}")))?;
}
```

Then in `service_graph.rs:210-214`:

````suggestion
    let models = Arc::new(ModelOps::new(ModelDeps {
        core: Arc::clone(&core),
        // The shared proxy runtime — the only registry that knows what is
        // actually running, now that ServerOps starts models through it.
        runtime: Arc::clone(&runtime),
        gguf_parser,
    }));
````

Also worth checking whether `ServiceGraphParams::runner` and `helpers::find_handle` still have any caller after this — `runner` may drop out of the graph entirely.

File: crates/gglib-runtime/src/server_config.rs [hunk 2]
```diff
@@ -158,5 +176,245 @@ pub fn build_server_config(
...
+pub fn resolve_unified_config(input: &UnifiedServerConfig) -> ServerConfig {
+    build_server_config(
+        input.model_id,
+        input.model_name.clone(),
+        input.model_path.clone(),
+        input.globals.llama_base_port,
+        &input.model_tags,
+        input.resolved_options(),
+    )
+}
```
mmogr: `resolve_unified_config` has no callers outside this file's own test module. Every real surface — `serve.rs:112`, `servers.rs:201` — calls `UnifiedServerConfig::resolved_options()` and hands the flattened `ServerConfigOptions` to the process manager, which reaches `build_server_config` itself at `swap_state.rs:326`. This function is never on that path.

The doc comment says "This is the entry point every launch surface should reach for," which is currently false. Two ways out, both fine, but we should pick one before this hardens:

**(a) Delete it.** Removes ~10 lines of function plus ~180 lines of tests that only test the function against `build_server_config`. Also needs: the `use crate::unified_server_config::UnifiedServerConfig;` import at line 56, and the two test call sites at lines 240 and 342.

**(b) Make the claim true.** Three CLI handlers still call `build_server_config` directly with hand-assembled options and no cascade — `handlers/plan.rs:129`, `handlers/council/mod.rs:128`, `handlers/agent_chat/config.rs:192`. Routing those through `resolve_unified_config` would give it real callers and close a genuine remaining DRY gap. Those files are outside this PR, so it'd be a follow-up.

I'd lean (a) now and (b) as a tracked follow-up — an abstraction that documents itself as central while being unreachable is worse than no abstraction.

crates/gglib-runtime/src/server_config.rs
```diff
@@ -26,6 +26,23 @@
...
+//! |----------|---------|----------|
+//! | [`resolve_unified_config`] | "which tier wins, *then* which flags?" | the caller holds a [`UnifiedServerConfig`] — CLI, GUI, proxy |
+//! | [`build_server_config`] | "which flags?" | the caller has already flattened its options |
+//!
+//! [`resolve_unified_config`] is a *cascade* layer: it resolves tier
+//! precedence and then hands the flattened result straight to
+//! [`build_server_config`], which remains the sole translator from options to
+//! llama-server arguments. It deliberately does not re-derive jinja, reasoning
+//! format, MTP or KV cache types — duplicating those resolvers is precisely
+//! the drift this module exists to prevent.
```
mmogr: This table presents both entry points as live options. Until #2a is resolved, the top row describes a function nothing calls.

If you take option (a), this is the surgical doc fix — replaces lines **30-45**:

````suggestion
//! ## One translator
//!
//! [`build_server_config`] is the sole translator from options to
//! llama-server arguments. Callers that carry a
//! [`UnifiedServerConfig`](crate::unified_server_config::UnifiedServerConfig)
//! flatten its three tiers with `resolved_options()` first; the process
//! manager then calls this function once, at spawn.
//!
//! Jinja, reasoning format, MTP and KV cache types are resolved here and
//! nowhere else — duplicating those resolvers is precisely the drift this
//! module exists to prevent.
````

File: crates/gglib-runtime/README.md [hunk 4]
```diff
@@ -108,18 +109,49 @@ See the [Architecture Overview](../../README.md#architecture) for the complete d
...
+let config = resolve_unified_config(&UnifiedServerConfig {
+    model_id,
+    model_name,
+    model_path,
+    model_tags: model.tags.clone(),
+    explicit: ServerConfigOptions { mlock: Some(true), ..Default::default() },
+    globals: GlobalDefaults::default(),
+    pinned: false,
+});
+```
```
mmogr: >This section documents `resolve_unified_config` as the cascade entry point every surface uses, with a runnable example. No surface calls it. Whichever way #2a lands, this needs to change — right now it will send the next contributor down a path that doesn't exist.

Surgical doc fix if you delete the function — replaces lines **125-149**:

````suggestion
| Layer | Entry point | Answers |
|-------|-------------|---------|
| Cascade | `UnifiedServerConfig::resolved_options` | which tier wins |
| Translation | `build_server_config` | which flags |

`UnifiedServerConfig` carries the three tiers — explicit overrides, curated
model defaults, global defaults — and `resolved_options()` flattens them into
a single `ServerConfigOptions`. The process manager then calls
`build_server_config` once at spawn. Capability detection therefore lives in
exactly one place; adding a resolver to `build_server_config` reaches every
surface automatically.

```rust,ignore
use gglib_runtime::{UnifiedServerConfig, GlobalDefaults};
use gglib_core::server_config::ServerConfigOptions;

let options = UnifiedServerConfig {
    model_id,
    model_name,
    model_path,
    model_tags: model.tags.clone(),
    explicit: ServerConfigOptions { mlock: Some(true), ..Default::default() },
    globals: GlobalDefaults::default(),
}
.resolved_options();
```
````

*(Note: this suggestion also drops `pinned` from the example — see Finding 3.)*

File: crates/gglib-proxy/README.md [hunk 5]
```diff
@@ -255,9 +255,23 @@ flags — no extra configuration required:
 
 Tags are detected automatically at model import time from GGUF metadata and
 stored in the model catalog. This parity is architecturally enforced: the proxy,
-GUI, and CLI all route through `build_server_config` in `gglib-runtime`, so any
-model that works correctly when started from the GUI or CLI will behave
-identically when auto-started by the proxy.
+GUI, and CLI all route through `resolve_unified_config` and its single
+translator `build_server_config` in `gglib-runtime`, so any model that works
+correctly when started from the GUI or CLI will behave identically when
+auto-started by the proxy.
```
mmogr:
````suggestion
Tags are detected automatically at model import time from GGUF metadata and
stored in the model catalog. This parity is architecturally enforced: the proxy,
GUI, and CLI all route through the single translator `build_server_config` in
`gglib-runtime`, so any model that works correctly when started from the GUI or
CLI will behave identically when auto-started by the proxy.
````

File: crates/gglib-runtime/src/unified_server_config.rs [hunk 6]
```diff
@@ -0,0 +1,491 @@
...
+    pub globals: GlobalDefaults,
+
+    /// Whether the HTTP surface should refuse every model but this one.
+    ///
+    /// `true` for `gglib serve <model>`, which exists to give single-model
+    /// clients (VS Code Copilot's BYOK endpoint, for one) a fixed endpoint.
+    /// `false` for `gglib proxy` and the GUI, where switching models is the
+    /// point.
+    pub pinned: bool,
+}
```
mmogr: Five of the eight fields on this struct are never read in production.

`model_id`, `model_name`, `model_path` and `model_tags` are populated by both real callers (`serve.rs:75-78`, `servers.rs:172-175`) and read only by `resolve_unified_config` (`server_config.rs:206-208`) — which has no callers at all (see 2). Neither `resolved_options()` nor `to_proxy_config()` touches them.

`pinned` (line 148) is stricter still: **nothing reads it anywhere**, not even a test. `serve.rs:109` sets `true`, `servers.rs:197` sets `false`, `server_config.rs:247/349` set `false`, and no code branches on it. Real pinning is driven entirely by `StandaloneProxyParams.pinned: Option<PinnedModel>` (`params.rs:159`) → `ProcessManager::new_pinned`.

The field is actively misleading — its doc comment describes behaviour it does not cause. At minimum `pinned` should go. Whether the four identity fields go depends on how 2 lands: keep them if we route the three direct `build_server_config` callers through the cascade, drop them if we delete it.

### Target state (not directly committable — `pinned` removal touches 5 files)

Removing `pinned` requires coordinated edits at:

- `crates/gglib-runtime/src/unified_server_config.rs` — field at :148 + doc at :142-147, test fixture at :225
- `crates/gglib-runtime/src/server_config.rs` — :247, :349
- `crates/gglib-cli/src/handlers/inference/serve.rs` — :109, :191
- `crates/gglib-app-services/src/servers.rs` — :195-197

The struct docs at `unified_server_config.rs:16-22` should also stop listing pinning as a tier concern.

crates/gglib-runtime/src/process/manager.rs
mmogr: After this PR, ProcessStrategy::Concurrent has no production callers left. The GUI was its only user, and service_graph.rs now builds a single-swap manager for every surface. The remaining new_concurrent call sites are all tests.
We should either delete it entirely (which is the DRY-consistent choice), or at least correct the docs to stop advertising it as the GUI strategy. Here is the doc fix if we keep it for tests:
```
/// Supports two strategies:
/// - **SingleSwap**: One model at a time, auto-swap (proxy, GUI, and
///   `gglib serve` in pinned mode) — use `new_single_swap` / `new_pinned`
/// - **Concurrent**: Multiple models at once — `new_concurrent`. No production
///   caller since every surface moved onto the shared single-swap manager;
///   retained for tests only.
```

mmogr:
```suggestion
/// Create a new `ProcessManager` with Concurrent strategy.
    ///
    /// No production caller: every surface now shares one `SingleSwap` manager
    /// (see `build_service_graph`). Retained for tests.
```

mmogr (commented):
Reviewed the full epic branch: build and clippy are clean, 353 tests pass, and I verified gglib serve end to end against a live model — the spawned llama-server carries every flag the epic promised (--host 127.0.0.1 --metrics --parallel 1 --slot-save-path --cache-ram --cache-type-k/v plus tag-driven --jinja/--reasoning-format/MTP), /v1/models advertises only the pinned model, a foreign model returns 404 pinned_model_mismatch, /v1/proxy/status serves the dashboard, and completions work. The core goal — serve as a mode of the proxy rather than a second stack — is genuinely achieved, and the structural wins are real (manager.rs 864→652, tauri/bootstrap.rs 604→384, four copies of the ops wiring collapsed into build_service_graph).

What's incomplete is the edges. One live regression: ModelOps still reads the ProcessRunner registry that ServerOps abandoned, which kills the remove force guard (1). One abstraction that documents itself as central while being unreachable (resolve_unified_config, 2) plus the write-only fields that exist to serve it (3). Dead strategy code the cutover left behind (4). And two acceptance criteria not fully met: GUI cache parity (5) and the KV-cache-persistence test #633 asked for (10).

Suggested order: 1 first (regression), then 2/3/4 as one cleanup, then 6 and 9 (UX), then 10.

PR #674, opened 2026-07-30
test: scratch PR to exercise inheritance live

### LINKED ISSUE #649: gglib proxy dashboard drifts/repeats on a narrow terminal
## Summary

`gglib proxy dashboard` redraws in place by moving the cursor up N rows and clearing everything below before printing the next frame. N is computed as `frame.lines().count()` — the number of `n`-delimited *logical* lines in the generated text, not the number of *physical* terminal rows those lines actually occupy.

When a terminal is narrow enough that a generated line wraps onto a second physical row, the redraw undercounts how far to move the cursor up, so each tick redraws slightly lower than the last. The dashboard then prints a growing trail of frames instead of redrawing over the previous one — exactly what's shown below on a narrow terminal:

```
gglib proxy dashboard — http://127.0.0.1:8080/v1/proxy/status/stream
gglib proxy dashboard — http://127.0.0.1:8080/v1/proxy/status/stream
gglib proxy dashboard — http://127.0.0.1:8080/v1/proxy/status/stream
... (repeats many times, scrolling down)
```

## Root cause

`crates/gglib-cli/src/handlers/proxy_dashboard.rs` already truncates two fields with genuinely unbounded content (the `/slots` failure reason and cache warnings) to fit within `term_width`, anticipating exactly this class of bug. But every other line — including the `gglib proxy dashboard — {url}` header itself — relies on fixed-width formatting tuned for an ~80-column terminal and is never checked against the live `term_width`. On a sufficiently narrow terminal, even the header line alone is long enough to wrap, reproducing the bug regardless of the existing truncation guards.

## Fix

Replace the naive `frame.lines().count()` redraw math with a helper that computes physical row count via ceiling-division against the live terminal width for every line, not just the two truncated fields — mirroring how the terminal itself auto-wraps in cooked/normal mode.

mmogr:
Closes #649

Throwaway PR referencing a real, fully-labelled closed issue to verify inheritance live. Will close without merging once confirmed.

### DIFF
--- a/crates/gglib-hf/src/lib.rs
+++ b/crates/gglib-hf/src/lib.rs
@@ -29,3 +29,4 @@ pub use config::HfClientConfig;
 use mockall as _;
 #[cfg(test)]
 use tokio_test as _;
+// scratch test: inheritance from linked issue


mmogr:
Verified live: component/priority/type correctly inherited from linked issue #649, size: correctly NOT inherited (Label Check failed on exactly and only size: as expected). Closing — throwaway test PR.

PR #694, opened 2026-08-04
Run the proxy as a background service from the system tray

mmogr:
## Why

The proxy has always been a feature you switch on from inside gglib, which makes it awkward as an endpoint for other clients: something has to be open, and something has to have been clicked. Pointing VS Code Copilot at it means keeping a terminal or a window around for a service that should just be there.

This makes the proxy a background service: it starts with the app, the window closes to the tray without stopping it, and the app itself can start at login.

### Why a tray, and not a container

The original idea was a Podman container so gglib could always be running. That is the wrong tool for this codebase. gglib is not a router to remote APIs — it spawns and swaps local `llama-server` processes (`crates/gglib-runtime/src/proxy/supervisor.rs`). Podman on macOS runs a Linux VM, and Metal cannot cross that boundary, so a macOS container means CPU-only inference. Cross-platform parity bought that way means making the Mac as slow as the worst platform.

Tauri's tray is cross-platform in the way that actually matters here — macOS menu bar, Windows notification area, Linux AppIndicator — and native GPU acceleration (Metal, CUDA, Vulkan) is untouched because inference never leaves the host.

A Containerfile still makes sense for headless Linux servers, where GPU passthrough works. That is not this PR.

## What changed

**Three settings** (`crates/gglib-core/src/settings.rs`) — `proxy_autostart`, `close_to_tray`, `start_at_login`. `settings_kv` stores settings as key-value JSON, so no migration.

**Autostart** (`src-tauri/src/autostart.rs`) runs after the window exists, so a slow or failing start delays nothing. Proxy startup goes through `ProxyOps::ensure_running`, which is idempotent and reads the saved `proxy_port` — a standing `gglib proxy` on that port is left alone rather than collecting a bind conflict at every launch. Login-item registration uses `tauri-plugin-autostart` (macOS login item, Windows `Run` key, XDG autostart entry) and is applied on every launch rather than only on change, so an entry removed outside gglib is restored. Both operations are best-effort and log on failure; refusing to open because a login item could not be registered would be worse than opening and saying so.

**Tray** (`src-tauri/src/tray/`) — the icon reflects the **proxy**, not the app. gglib being open says nothing about whether anything is being served, and that is the question the tray exists to answer. Left-click opens a popover with the endpoint, live connections and slots, and start/stop; right-click gives the same plus open, preferences and quit.

**Shared components** (`src/components/proxy/`) — the popover shows the same figures as the in-app dashboard, so the display logic was pulled out before the third consumer existed rather than after. `ProxyControl` and `ProxyDashboardModal` now render through it, and `ProxyControl` drops from 286 to 255 lines — under the project's 300-line budget for the first time. Connections and slots are exported as separate sections as well as a composed pair, because the modal renders its cache panels between the two; the dashboard's section order is unchanged.

## Design notes

**No new IPC commands.** The popover is a second Vite entry (`tray.html`) that reaches the embedded API through the existing HTTP transport, using the already-allowlisted `get_embedded_api_info` for the bearer token. Window-level actions stay on the native tray menu, handled in Rust — that is what keeps `check-tauri-commands.sh` and `check-frontend-ipc.sh` untouched, and it means the panel tests run under jsdom with no Tauri mocking. It builds to 2.3 kB against the main entry's 276 kB.

**Tray actions go direct, not through the frontend.** The application menu drives the proxy by emitting an event the webview turns into an HTTP call. The tray cannot assume a webview is loaded — it is reachable with the window hidden, and autostart runs before one exists. So `src-tauri/src/proxy_actions.rs` calls the same `ProxyOps` the Axum server uses, then does by hand what the HTTP handler would have: publish state and broadcast the lifecycle event. Skipping that broadcast is what would leave the UI showing a stopped proxy while it serves.

**Cross-platform state sync.** `state_sync` drives both surfaces. Rather than build one combined state object and leave half unread off macOS, each surface gathers what it needs — `AppMenu` and `MenuState` are now macOS-only, and Linux and Windows build warning-free.

**Quit now asks.** With close-to-tray on, quitting is the one action that takes the endpoint away from whatever is still pointed at it. Both quit paths keep the existing hardened shutdown and its watchdog.

## Known limitation

Linux's AppIndicator delivers no click events, so the panel opens from the "Proxy Panel" menu item rather than a left-click. Every action is on the menu on all platforms for exactly this reason — the click gesture is a shortcut, never the only route to a feature. This is documented in `src-tauri/src/tray/README.md`.

The tray tooltip reports the port but not a live request count: the connection registry lives inside the running proxy and is not reachable from the supervisor without threading it through three crates. The panel shows live connections via the existing dashboard SSE stream, which is where that information belongs.

## Verification

Run on Linux (this environment), with `libwebkit2gtk-4.1-dev` and `libayatana-appindicator3-dev` installed:

- `cargo test` on every touched crate — **976 passed, 0 failed**, including 4 new unit tests for the pure icon/tooltip derivation
- `cargo clippy --all-targets` on every touched crate — **0 warnings**, with the workspace's `pedantic` + `nursery` lints and `unsafe_code = "deny"`
- `npm run test:run` — **672 passed**, including 26 new tests across the shared components, the settings hook, and the tray panel
- `npx tsc --noEmit` and `npm run lint` — clean
- `npm run build` — both entries emit (`tray.html` at 2.3 kB, main at 276 kB)
- CI gates: `check_boundaries.sh` (runs `check_readmes.sh --strict`), `check-tauri-commands.sh`, `check-frontend-ipc.sh`, `check_transport_branching.sh` — all pass

**CLI parity**, exercised against a real database rather than inspected:

```
$ gglib config settings set --proxy-autostart true --close-to-tray true
✓ Settings updated successfully:
  close-to-tray    true
  proxy-autostart  true

$ gglib config settings show
  close-to-tray               true
  proxy-autostart             true
  start-at-login              None
```

GUI and Web UI parity is structural: `gglib web` serves the same React build, so the settings group renders identically in both.

### Not verified here

The tray icon, popover positioning and close-to-tray behaviour have **not been exercised on a running desktop** — this environment is headless, so the Tauri app compiles and its logic is unit-tested but never rendered. macOS specifics in particular are unverified: template-image rendering in the menu bar, `RunEvent::Reopen` on dock click, and the login item. Worth a manual pass on macOS before merge:

1. Enable autostart + close-to-tray, restart → menu bar icon active, `curl 127.0.0.1:8080/v1/models` returns the library with no terminal open
2. Close the window → app alive, proxy serving; click the icon → panel shows the endpoint and live connections
3. Request from VS Code Copilot → connection appears in the panel, clears on completion
4. Quit from the tray → confirmation, then `pgrep -fl llama-server` shows no orphans

## Note on file complexity

`scripts/check_file_complexity.sh` fails on `main` with 30 pre-existing violations and is not wired into `ci.yml`. Every file added here is well under 300 lines, and the extraction brought `ProxyControl.tsx` under it; the pre-existing violations are untouched as out of scope.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FFaoaNfWJxf8WC6aV2Ptuk)_


mmogr:
Superseded by #695 after renaming the branch to `feat(proxy)/tray-background-service` (author identity on all commits was also corrected to mmogr).

PR #736, opened 2026-08-07
fix(frontend): make InferenceParametersForm tell the truth about what an empty field inherits

### LINKED ISSUE #735: fix(frontend): InferenceParametersForm labels aren't associated, and its default caption disagrees with the rest of settings
Split out of #733 / #734, which fixed the equivalent problems in `SettingsModal/fields/` but deliberately stopped short of this component — it also renders in `ModelEditForm` and `ServeModal`, so it's a three-surface change with a design call in it.

`src/components/InferenceParametersForm/InferenceParametersForm.tsx`

## 1. Labels aren't programmatically associated

Both render helpers emit a bare `<label>` with no `htmlFor`, and the `Input` next to it has no `id`:

- `renderNumberInput` — line 58
- `renderSlider` — line 111

So the label isn't announced with the control, and nothing in this form is reachable via `getByLabelText` — which is why the new `SettingsFields.test.tsx` covers the five `SettingField` inputs and none of these.

## 2. Two phrasings for one idea, side by side

In the same Advanced Settings panel:

- Context Size etc. render `Default: 4096`, always visible.
- Top K renders italic `Using default (40)`, and only while the field is unset (`{!isSet && …}`, line 88).

The sliders are a third variant again: `1.00 (default)` appended to the value readout (line 126).

The design call is whether the caption should become always-visible like `SettingField`'s hint. Arguments both ways — this form has a reset (X) button that already signals "a value is set", which the settings fields don't, so the disappearing caption is not obviously wrong here. Worth deciding deliberately rather than by drift.

## 3. Also worth folding in: a Rust↔TS contract test

Noted while doing #734. `src/constants/settingsDefaults.ts` mirrors `Settings::with_defaults()` and `validate_settings()` in `crates/gglib-core/src/settings.rs` by hand and documents the coupling, but nothing enforces it — a Rust-side change to a default or a range goes silently out of sync with what the GUI promises the user.

`tests/ts/contracts/` is the place for it (one test file and a `TODO` README today). The same applies to this component's hardcoded `'0.7'` / `'0.95'` / `'40'` / `'2048'` hints, which have the same relationship to `validate_inference_config`.

mmogr:
Closes #735. **Stacked on #734** — based on `fix(frontend)/settings-default-placeholders`, so review only the eight commits here. GitHub retargets this to `main` when #734 merges.

#735 was filed as cosmetic: missing `htmlFor`, and `Using default (40)` being a second phrasing for #734's `Default: N`. Planning it turned up three substantive problems, one of which changes the whole approach.

## 1. The caption was wrong on two of its three surfaces

Sampling resolves through a ladder — request → profile → per-model → global settings → floor. The form renders at three different rungs:

| Surface | Rung | What an empty field really falls through to |
|---|---|---|
| Settings → Global Inference Parameter Defaults | global settings | the floor — correct here |
| `ModelEditForm` | per-model | **the global setting** |
| `ServeModal` | per-request | profile → model → global, then the floor |

All three showed the floor's number and called it the default. A user with a global temperature of 1.2 was told their model would use 0.7. `ServeModal` contradicted itself on screen: its own prose said *"leave empty to use model or global defaults"* directly above captions naming hardcoded values.

So #735's "should the caption become always-visible?" resolves to **yes, but only after making it true** — otherwise it just makes a wrong statement more prominent.

## 2. Max Tokens claimed a default that deliberately does not exist

`with_hardcoded_defaults()` sets `max_tokens: None`, and spends twenty lines of docstring on why: resolution force-writes every `Some` field, so a fallback there would silently cap *every* request that didn't name its own. Unset, llama-server generates until a stop token or the context limit.

The form hinted `2048`. Someone who believed a 2048 cap was in force and typed 8192 to raise it was imposing the first cap on their own requests — the exact truncation the Rust docstring exists to prevent.

## 3. Repeat Penalty could produce a value the backend rejects

Slider `min = 0`; `validate_inference_config` rejects `repeat_penalty <= 0.0`. Drag to the left stop, save, error — with nothing on screen suggesting that end of the track was out of bounds.

## Approach: ask, don't re-derive

`resolve_layers_with_sources` already returns the resolved value *and* its provenance in one pass, and its docstring is an explicit warning against the obvious alternative:

> a second function that re-derived the rules would eventually explain a decision the resolution did not take, which is exactly what the `describe_provenance` helper this replaced had already started doing

So none of the coupling rules (the temperature-coupled trio, the reasoning floor) are reimplemented in TypeScript. The model surfaces read `GET /api/models/:id/explain` through the existing `useSamplingExplanation` hook and format it with the existing `describeSource` — the same source of truth, and the same wording, as the inspector's Sampling section a few hundred pixels away. The settings surface keeps a static `Default: 0.7`, which is correct precisely there.

A required `fallback` prop names the rung each surface edits — required rather than defaulted, so a fourth surface has to say what it is instead of inheriting a caption describing someone else's layer.

Three consequences of the explanation describing *saved* state:

- **The placeholder and slider thumb now sit on the value that will actually apply**, so the number on the row agrees with the caption under it. This also retires the readout's third phrasing of "default".
- **A value owned by the layer this form edits gets no caption** while the field is empty — the user just cleared it, and the saved provenance is one save out of date.
- **Loading claims nothing**; a failed or absent explain route says `Resolution unavailable` rather than quietly falling back to a hardcoded number.

## Commits

| | |
|---|---|
| `b366fba` | **fix** — `htmlFor`/`id`/`aria-describedby`. Nothing here was announced with its label or reachable by `getByLabelText`, which is why everything below was untestable. |
| `9fc419c` | **refactor** — move `samplingProvenance.ts` to `src/utils/`; pure move. |
| `d768186` | **refactor** — `src/constants/inferenceDefaults.ts`, one `{default, min, max, step}` per parameter naming the Rust symbol it mirrors. Behaviour-neutral, so the next two are one-line diffs. |
| `207ceea` | **fix** — Max Tokens claims no default. |
| `b16fa13` | **fix** — Repeat Penalty starts at 0.05. |
| `5882ee3` | **feat** — the substance: resolved value + provenance per surface. |
| `9e263bb` | **test** — contract test parsing the Rust validators. |
| `6300858` | **test** — 33 tests over the form. |

## Testing

- 691/691 pass (57 new). `tsc --noEmit`, `eslint src/`, `check_readmes.sh`, `check_file_complexity.sh` all clean for the files touched.
- **The contract test bites**: reverting `207ceea` and `b16fa13` fails it twice, naming `repeatPenalty` and `maxTokens` rather than reporting a generic mismatch.
- **The component tests bite**: making the caption always describe the floor fails six of them.
- Worth flagging from writing the contract test — an early draft scanned forward from a bare field name, read Presence Penalty's range out of the block below, and **declared the Repeat Penalty bug clean**. Guards are now anchored on their subject (`config.top_p`) and bounded to a single `if` block, and that trap is recorded in a comment.

Manual, all three surfaces: settings reads `Default: 0.7` / `Default: no limit`; with a global temperature of 1.2 set, the model editor reads `Resolves to 1.2 — global settings` where it used to say `0.70 (default)`; every row cross-checks against the read-mode Sampling section, since both now come from the same endpoint.

## Follow-up

`tests/ts/contracts/README.md` was a `TODO` and is now filled in. One thing deliberately left: the `SamplingProvenanceSection` and this form now fetch the same explanation independently within the inspector panel — hoisting that to one fetch is a reasonable cleanup, but not one to bundle into a bugfix.

---
_Generated by [Claude Code](https://claude.ai/code/session_016M4zS1W1T91NTDguTW5Pjk)_

### DIFF
--- a/src/components/InferenceParametersForm/InferenceParametersForm.tsx
+++ b/src/components/InferenceParametersForm/InferenceParametersForm.tsx
@@ -59,7 +59,9 @@ export const InferenceParametersForm: FC<InferenceParametersFormProps> = ({
   const renderNumberInput = (field: SamplingParamKey) => {
     const { default: fallback, min, max, step } = INFERENCE_PARAMS[field];
     const label = PARAM_LABELS[field];
-    const defaultHint = formatParamValue(field, fallback);
+    // A parameter the floor leaves unset has no number to offer — no
+    // placeholder to type over, and nothing to name in the caption.
+    const defaultHint = fallback === null ? null : formatParamValue(field, fallback);
     const currentValue = config[field];
     const isSet = currentValue !== undefined && currentValue !== null;
     const inputId = paramId(field);
@@ -76,7 +78,7 @@ export const InferenceParametersForm: FC<InferenceParametersFormProps> = ({
               const val = e.target.value;
               updateField(field, val === '' ? undefined : Number(val));
             }}
-            placeholder={defaultHint}
+            placeholder={defaultHint ?? undefined}
             min={min}
             max={max}
             step={step}
@@ -100,7 +102,9 @@ export const InferenceParametersForm: FC<InferenceParametersFormProps> = ({
         </div>
         {!isSet && (
           <span id={paramCaptionId(field)} className="text-xs text-text-muted italic">
-            Using default ({defaultHint})
+            {defaultHint === null
+              ? 'No limit — generates until the context is full'
+              : `Using default (${defaultHint})`}
           </span>
         )}
       </div>
@@ -112,7 +116,9 @@ export const InferenceParametersForm: FC<InferenceParametersFormProps> = ({
     const label = PARAM_LABELS[field];
     const currentValue = config[field];
     const isSet = currentValue !== undefined && currentValue !== null;
-    const displayValue = isSet ? currentValue : fallback;
+    // Every slider parameter has a floor; `?? min` only satisfies the type,
+    // which is nullable for Max Tokens' sake — and that one is a number input.
+    const displayValue = isSet ? currentValue : (fallback ?? min);
     const inputId = paramId(field);
 
     return (
--- a/src/components/InferenceParametersForm/InferenceParametersForm.tsx
+++ b/src/components/InferenceParametersForm/InferenceParametersForm.tsx
@@ -1,6 +1,8 @@
 import { FC, useCallback, useMemo } from 'react';
 import { X } from 'lucide-react';
-import type { InferenceConfig } from '../../types';
+import type { InferenceConfig, SamplingParamKey } from '../../types';
+import { INFERENCE_PARAMS } from '../../constants/inferenceDefaults';
+import { PARAM_LABELS, formatParamValue } from '../../utils/samplingProvenance';
 import { Input } from '../ui/Input';
 import { Icon } from '../ui/Icon';
 import { IconButton } from '../ui/IconButton';
@@ -19,10 +21,10 @@ interface InferenceParametersFormProps {
  * spelled out here rather than imported: this form is a top-level component
  * and has no business depending on the settings modal's internals.
  */
-const paramId = (field: keyof InferenceConfig) => `inference-param-${field}`;
+const paramId = (field: SamplingParamKey) => `inference-param-${field}`;
 
 /** DOM id for the caption below a parameter, referenced by `aria-describedby`. */
-const paramCaptionId = (field: keyof InferenceConfig) => `${paramId(field)}-description`;
+const paramCaptionId = (field: SamplingParamKey) => `${paramId(field)}-description`;
 
 /**
  * Tristate inference parameters form.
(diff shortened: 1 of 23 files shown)


mmogr:
Note on the check list above: `ci.yml` triggers on `pull_request: branches: [main]`, so only the label jobs run while this targets #734's branch. The full suite will run automatically once #734 merges and GitHub retargets this to `main` — the green tick here is not evidence the build passes.

Run locally against this branch in the meantime:

| | |
|---|---|
| `vitest run` | 691/691 pass (57 new) |
| `tsc --noEmit` | clean |
| `eslint src/` | clean |
| `check_readmes.sh` | pass |
| `check_file_complexity.sh` | 20 violations, all pre-existing on `main`, none in files touched here |

Rust is untouched, so `cargo fmt`/`clippy`/`test` are unaffected — though `settingsBounds.test.ts` now reads `settings.rs` and `inference.rs`, which means a Rust-side change to a default or a validation range will surface as a **frontend** test failure. That is the intent, but it is a new coupling worth knowing about before it surprises someone.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #771, opened 2026-08-10
chore(cli): remove the assistant-ui command

mmogr:
## What

Removes `gglib config assistant-ui` — the command that shelled out to npm to manage the frontend's own packages.

Stacked on `chore(repo)/dead-code-cleanup` (#770).

## Why

It is a build-time concern wearing a runtime command's clothes, and it had stopped working besides:

**It only functioned from a repo checkout.** `handle_status` checks `PathBuf::from("node_modules")` — a relative path — so in an installed binary it inspected whatever directory the user happened to run from. `handle_install` ran a bare `npm install` in the process CWD.

**Its verification could not succeed.** `verify_installation` checked for `@assistant-ui/react-ai-sdk`, which the project no longer depends on.

**Nothing called it.** No routes, no GUI, no setup wizard — only its own CLI dispatch arm.

## What stays

`@assistant-ui/react` is untouched and remains a dependency. About ten source files import it — the tool usage badge, the tool details modal, the generic tool UI, the message content-part helpers — and the frontend build is verified green here. Updating that package is what npm is for.

## Docs

Four READMEs described this module, one of them as a *"terminal-based interactive chat UI for llama-cli"*, which it never was. Those entries, their badge rows and the architecture-diagram box go with the code.

## Test plan

- `cargo test --workspace` — green
- `cargo clippy --workspace --all-targets` — zero warnings
- `gglib config --help` no longer lists the subcommand; the other eight are unchanged
- `npm run build` — green, confirming the library is untouched


mmogr:
Heads-up on a stack that touches the frontend gates: #776 makes `tests/ts/**` type-checked for the first time (`tsc -b` over a new `tsconfig.test.json`), and #774 adds a CI job that runs eslint and tsc.

This PR adds `tests/ts/components/DiagnosticsPanel.test.tsx` and `RecommendedModel.test.tsx`, so once that stack lands they will need to type-check. Since this branch already needs a conflict-resolution pass against main, worth folding in at the same time.

No file overlap between the two: the only config-adjacent file here is `.github/workflows/badges.yml`, which has no `setup-node` step and is untouched by the stack.

PR #775, opened 2026-08-11
chore(cargo): optimize dev deps, resolver 3, per-crate tokio, one version source

mmogr:
**Stack 2/4** — base `ci(workflows)/lockfile-cache-and-gates` (#774). Retarget to `main` once that merges.

### `[profile.dev.package."*"] opt-level = 3`

Optimize dependencies in debug builds; our own code stays at 0 so incremental rebuilds are unaffected. Deps compile once and stay optimized — which matters here, where debug builds hash multi-GB GGUFs through `sha2` and stream downloads through `reqwest`/`rustls`. Same for `[profile.test]`.

### Resolver 2 → 3

Edition-2024 packages default to resolver 3, but a virtual manifest has no edition to infer from and must state it. The pinned 1.97.1 toolchain is well past the 1.84 required. `Cargo.lock` is unchanged by the switch.

### Per-crate tokio features

The workspace entry inherited `features = ["full"]` into all 14 consumers; it now declares **no features at all**, because features listed in `[workspace.dependencies]` are additive and unconditional and would defeat the per-crate sets. `gglib-proxy` also bypassed the workspace entry with its own full-featured copy.

Sets were derived by scanning for tokio-specific paths, then converged with `cargo check -p <crate> --all-targets`. Three seeds were grep false positives, dropped after reading the call sites:

| seed | reality |
|---|---|
| `io-std` in gglib-cli, gglib-runtime | matched `std::io::stdout` |
| `process` in gglib-cli, gglib-mcp, src-tauri | matched `std::process::Command` |

`gglib-hf` and `gglib-db` got a `[dev-dependencies]` entry instead — only their inline `#[tokio::test]` blocks need `macros`/`rt`.

**Setting expectations:** this does *not* shorten a full workspace build. Cargo unifies features across the graph and `gglib-download`/`gglib-runtime` between them already need most of `full`. The measurable effect is `parking_lot` leaving the tokio build, plus genuinely cheap single-crate builds and manifests that state what they use.

### One version source

`Makefile` exported `CARGO_PROFILE_RELEASE_LTO`/`_CODEGEN_UNITS`, duplicating `[profile.release]` with identical values — and since env vars win, a later manifest edit would have been silently ignored under `make`. Deleted.

`"version"` is removed from `src-tauri/tauri.conf.json`, leaving `Cargo.toml` the only source. Tauri falls back to the crate version when absent — verified in the vendored sources, not just the docs:

- `tauri-codegen/src/context.rs:273` emits `env!("CARGO_PKG_VERSION")` for `PackageInfo`
- `tauri-winres` defaults `FileVersion`/`ProductVersion` to the same, so Windows resources are unaffected

`sync_versions.py` and `bump-version.yml` no longer patch or verify the key. The workflow now **fails if it comes back**, since a re-added key would silently pin the metadata again.

Also fixes `sync_versions.py` writing JSON with `ensure_ascii=True`, which rewrote the em dash in `package.json`'s `//overrides` note to an escape on every run.

### Verification

- `cargo check --workspace --all-targets` clean; `cargo check -p <crate> --all-targets` clean for all 15 + `gglib-app`
- `cargo test --workspace` — **79 test binaries, 0 failures**. This matters: a missing tokio feature is a *runtime* panic (`time driver not enabled`), not a compile error.
- `cargo metadata --locked` clean for both manifests
- release build → `gglib 0.12.2`, confirming the version fallback
- `python3 scripts/sync_versions.py` → no drift

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -131,14 +131,24 @@ jobs:
         run: |
           VERSION="${{ steps.validate.outputs.version }}"
           
-          # Check that key files were updated
-          for file in package.json src-tauri/tauri.conf.json; do
+          # Check that key files were updated. src-tauri/tauri.conf.json is not
+          # in this list: it no longer carries a "version" key, because Tauri
+          # falls back to the Cargo.toml version when the key is absent.
+          for file in package.json; do
             if ! grep -q "\"version\": \"$VERSION\"" "$file"; then
               echo "::error::$file was not updated correctly"
               exit 1
             fi
             echo "✓ Verified $file"
           done
+
+          # Guard the fallback: a re-added "version" key would silently pin the
+          # app metadata and reintroduce the drift this removal eliminated.
+          if grep -q '"version"' src-tauri/tauri.conf.json; then
+            echo "::error::src-tauri/tauri.conf.json must not declare a version; Tauri inherits it from Cargo.toml"
+            exit 1
+          fi
+          echo "✓ Verified src-tauri/tauri.conf.json inherits its version"
 
       - name: Create Pull Request
         id: create-pr
@@ -162,8 +172,8 @@ jobs:
             - ✓ `package.json`
             - ✓ `src-tauri/Cargo.toml`
             - ✓ `src-tauri/Cargo.lock`
-            - ✓ `src-tauri/tauri.conf.json`
             - ✓ All workspace crates (via inheritance)
+            - ✓ `src-tauri/tauri.conf.json` (via Tauri's Cargo.toml fallback)
             
             ### Next Steps
             Once this PR is merged, the [Release workflow](.github/workflows/release.yml) will automatically:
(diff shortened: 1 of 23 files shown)


mmogr:
**Ordering note — interacts with #773 ("Bump version to 0.13.0").**

#773 was opened at 13:01 today, after this stack was scoped, and it touches all four files this PR does: `Cargo.toml`, `Cargo.lock`, `package.json`, `src-tauri/tauri.conf.json`.

One real conflict, and it is trivial: #773 sets `"version": "0.13.0"` in `src-tauri/tauri.conf.json`; this PR **deletes that key**. Whichever order they merge, the resolution is to take the deletion — the whole point is that Tauri inherits the version from `Cargo.toml`. The other three files touch different lines and should auto-merge.

Worth being deliberate about the order:

- **#773 first** — resolve the one-line conflict here by keeping the deletion. Note that #773 was generated by the *old* `bump-version.yml`, so it still patches `tauri.conf.json`; this PR is what stops that happening again.
- **This stack first** — #773 needs regenerating, since the updated workflow no longer patches (and now actively rejects) that key.

Slight preference for merging #773 first: it is a one-line resolution here, versus re-running the bump workflow.

PR #779, opened 2026-08-11
fix(core,proxy): close the untrusted-sampler passthrough

mmogr:
Closes the untrusted-sampler passthrough: the one remaining way an external client could steer sampling past `trust_client_sampling: false`. First PR of the closed-loop measurement arc — this one protects the *consistency* the loop will depend on, before any tuning automation lands on top of it.

## What was wrong

The trust gate discards the client's sampling **layer**, but `resolve_sampling` only ever *inserts* the resolved patch into the body — nothing removed keys the ladder has no field for. Two consequences, one worse than the other:

1. **`frequency_penalty`** — an OpenAI-standard field llama.cpp supports, `presence_penalty`'s twin — was never modelled, so a client sending it reached llama-server on every request while every modelled twin was gated. ADR 0003 filed this follow-up as "decide deliberately rather than by omission".
2. **The general case.** Every sampler key gglib does not model rode through the same hole: `mirostat` (which replaces the entire configured truncation stack), `typical_p`, the XTC pair, `dry_sequence_breakers`, `repeat_last_n`, `min_keep`, and the `samplers` chain-order array itself. gglib's own values arrived intact, so the sampling audit saw no divergence — `/slots.params` echoes what was parsed, not what the chain did (ADR 0003 finding 7 states exactly this limit) — and the applied chain was something nobody configured. Production quality could silently diverge from benched quality with nothing reporting it.

## What changed

**`frequency_penalty` is modelled** through the full hierarchy: ladder, provenance, client extraction, explain surfaces (CLI table + GUI provenance panel), operator flags (`--frequency-penalty` on the shared sampling flags, `gglib config profile set` including `--unset`, `gglib model update` including dry-run diffs), and the GUI mirrors with the `settingsBounds` contract test pinning the TS bounds to the Rust guard. Deliberately **uncoupled** from the temperature trio even though the trio's rationale applies to it literally — because #746's cost applies literally too: coupling would cost a layer naming only a `frequency_penalty` its setting whenever a lower layer names a temperature, the default state of every `reasoning`-tagged model, and no shipped profile pairs the two, so coupling would protect nothing in exchange. No floor names it, per ADR 0003: llama.cpp's own 0.0 applies by silence.

**Unmodelled sampler keys are stripped from untrusted bodies.** `UNMODELLED_SAMPLER_KEYS` names the list; `strip_unmodelled_sampler_keys` runs beside the gate, before the fold. Scope is **taste, not function**: budgets, stops, `grammar`/`json_schema`/`response_format` and observation fields stay client-authoritative. `logit_bias` is a documented deliberate keep — per-token surgery with legitimate functional uses (banning a token); moving it deserves its own decision, not a sweep. Trusted clients keep every key: trusted means trusted.

**Both kinds of drop are recorded.** Stripped keys join `client_fields_discarded` beside the gate's discards, so the operator-facing record carries the whole of what an untrusted client asked for and did not get.

## Scope notes worth knowing before review

- **A disjointness test pins the boundary between gate and strip**: no modelled key may appear in the strip list, because stripping one would delete the client's value before the *trusted* path could read it. The test constructs a full `InferenceConfig` literal on purpose — a future field fails compilation there and forces its author to decide whether the strip list must shrink, which is exactly what modelling `frequency_penalty` required.
- **The sampling audit's compared set is untouched** (still the documented 7 fields), following the entropy-samplers precedent. `frequency_penalty` has no `/props` `UPSTREAM_DEFAULTS` entry either — that table holds *measured* values only, and no measurement of the pinned build was taken here; captions cite llama.cpp's documented 0.0 instead.
- **Not a sweep dimension.** Governing a parameter is today's decision; sweeping it is a tuning decision with its own evidence bar (#748's pattern).
- **Two pre-existing gaps fixed in passing**: `gglib config profile show` never gained rows for the three entropy-adaptive fields (a profile carrying one displayed as if it did not), and the `docs/sampling.md` parameter table was missing the same three rows.

## Verification

- Full workspace: `cargo test` — 79 groups, 0 failures. `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.
- TS: 816 tests across 60 files, 0 failures, including the extended `settingsBounds` contract (its guard parser learned that an inclusive range's bounds may be negative — the first negative lower bound in the validator).
- `check_boundaries.sh`, `check_readmes.sh`, `check_param_source_exhaustive.sh`, `check_transport_branching.sh`, `check-tauri-commands.sh` all pass; eslint clean.
- Four new pipeline tests: the strip + its record, the trusted no-op, functional keys surviving, the disjointness guard. The existing wire-key pairing and display-order contracts fired during development exactly as their doc comments promise, which is how the CLI explain table and DTO surfaces were caught.

### Not verified here

Live end-to-end against a running llama-server (this environment has no model installed): confirming a stripped `mirostat: 2` no longer changes `/slots`' reported chain, and that `--frequency-penalty 0.5` lands in `/slots.params.frequency_penalty`. Both are one-request checks worth doing before merge.


mmogr:
**Live hardware verification (Qwen3.5-4B Q8_0, Metal, llama-server b7ba604f) — passed after one fix.**

The check found a real hole beyond this PR's original scope: an untrusted client's `frequency_penalty: 0.9` reached `/slots` intact through the closed gate. The gate drops modelled fields from the client *layer*, but force-insert only overwrites keys resolution emits — and since ADR 0003, deferred fields emit nothing, so the client's body key survived. Pre-existing for every deferred field since the deferral; invisible to any test that names the field somewhere in the ladder. Fixed in `bf8c0deb` (the gate's drops now leave the body; regression + pipeline-level tests added in `f0b238b3`).

Verified after the fix, mid-generation via `/slots`:
- poisoned request (`mirostat: 2`, `typical_p: 0.5`, `min_keep: 5`, `frequency_penalty: 0.9`, `top_p: 0.3`, `dry_base: 3.5`, `seed: 42`) → server reports `frequency_penalty 0.0`, `mirostat 0`, `typical_p 1.0`, `min_keep 0`; recipe intact (`temperature 1.0`, `top_k 20`, `presence_penalty 1.5`).
- `gglib proxy --frequency-penalty 0.5` → `/slots` reports `0.5`, riding above the recipe.
- launch line carries zero sampler flags (ADR 0003 deletion holds on this build).

Operational note: the first re-test falsely failed because a stale auto-spawned daemon served the pre-fix binary — the same trap noted in the tune-benchmark runbook. Kill the daemon after rebuilding.

PR #782, opened 2026-08-11
feat(benchmark): eval gates — paired statistics and multi-pair A/A drift

mmogr:
> **Stacked on #780** (which stacks on #779). Base is `feat(core,db)/measured-defaults-origin`; retarget as the stack merges.

Third PR of the closed-loop arc, and the gate everything after it depends on: nothing may ever auto-apply a tuning decision the eval cannot resolve from its own noise. Two strengthenings, both prescribed by the instrument's own documentation.

## 1. Paired per-seed statistics — from data the report already stores

The two real arms run the same seeds on the same tasks, and the report's drill-down stores every `(task, seed)` cell — but the only comparison ever made was arm mean against arm mean. Pairing is what removes the eval's identical-arm spread from the comparison: the ceiling experiment (ADR 0004's postscript) resolved a +0.067 effect through noise wider than that *only* because it paired per run. The A/B eval now gets the same treatment.

`PairedEffect` carries matched-pair counts (unmeasured pairs dropped **and reported** — the `unmeasured_runs` rule at pair granularity), wins/losses/ties on `tool_match_score`, the mean paired delta, and a one-sided Wilcoxon signed-rank *p* — textbook construction: zeros dropped, average ranks with the tie correction, continuity-corrected normal approximation, and a refusal to speak below eight non-tied pairs rather than rendering a statistic the design cannot support (`EffectVerdict`'s own rule). Two tests pin the implementation against hand-computed bands: ten distinct all-positive deltas give W⁻ = 0 and p ≈ 0.0030 — an error of one rank, one correction term, or a dropped tail leaves the band — and magnitude-balanced deltas land on the null mean at chance.

**Stored, not derived** — a deliberate exception to the verdicts' derive-only pattern, because the verdicts re-derive from two floats in any language while this carries a rank test nobody should maintain twice (the GUI hand-mirrors the verdict math in `verdicts.ts`, and a TS Wilcoxon twin is exactly the drift-prone duplication the mirror comment warns about). `paired_effect()` still re-derives from the drill-down, so legacy reports gain the paired view retroactively.

## 2. Multi-pair A/A drift — the strengthening the docs always named

`EFFECT_NOISE_RATIO`'s doc has said from day one: "The honest way to strengthen this is more A/A pairs, not a bigger factor." Now there can be more pairs. `--replicate-pairs N` re-runs the raw arm on N derived disjoint seed sets; the drift floor becomes the **mean pairwise composite gap over all C(K+1, 2) pairs** among the primary and every replicate — same estimator (a gap, not an SD, so the 2.0 threshold keeps its calibration), more degrees of freedom. `EffectVerdict` now carries the pair count, because a verdict over one pair and one over six are not the same strength of claim.

Compatibility is pinned by test in both directions: pair 1's seed set is exactly the legacy derivation (a multi-pair run's first pair stays comparable with every earlier report), and a legacy report — populated `raw_replicate`, empty `raw_replicates` — reads exactly as it always did: one gap, one pair.

## Surfaces

- **CLI**: the report prints the paired line (`11W–4L–5T over 20 pairs, mean Δ +0.067; Wilcoxon one-sided p = 0.0099` — or an honest refusal below the minimum), names the gap count behind the drift, and describes a multi-pair A/A arm as what it is.
- **GUI**: the verdicts panel gains the paired block; `verdicts.ts`'s hand-mirrored `noiseFloor`/`effectVerdict` upgrade to the multi-pair mean with the pair count threaded through.
- **Types**: `AgenticEvalConfig.replicate_pairs`, `AgenticEvalReport.{raw_replicates, replicate_seed_sets, paired}` on both sides of the wire, all serde-defaulted so old configs and old reports deserialize unchanged.
- The axum handler needed nothing: it accepts `AgenticEvalConfig` as JSON, and the new field defaults server-side.

## Verification

- Full workspace `cargo test` green — 1023 tests in `gglib-core` (10 new: the Wilcoxon bands, pair accounting, unmeasured-pair drops, multi-pair drift arithmetic with hand-computed gaps, legacy single-pair equivalence, first-pair seed-set equality). `clippy --all-targets -- -D warnings` clean; fmt clean.
- TS: 816 tests green (the pinned verdict-mirror tests pass unchanged — single-pair behaviour is identical by construction), `tsc --noEmit` clean, eslint clean.

### Not verified here

A live multi-pair run (needs a model and GPU): `gglib benchmark agentic <model> --replicate-pairs 3` should print a drift floor labelled "mean over 3 pairwise gaps"… with 3 A/A arms it is 6 gaps among 4 raw runs. One eval's wall-clock worth of confirmation before merge.

## What this sets up

PR 4 (`gglib model tune --apply`) gates on exactly these numbers: a winner applies only when its margin clears the multi-pair drift and the paired comparison agrees with the direction. This PR makes those numbers exist.

### DIFF
--- a/crates/gglib-app-services/src/benchmark/agentic.rs
+++ b/crates/gglib-app-services/src/benchmark/agentic.rs
@@ -25,7 +25,7 @@ use gglib_core::domain::InferenceConfig;
 use gglib_core::domain::benchmark::agentic::{
     AgenticEvalConfig, AgenticEvalReport, AgenticTaskComparison, ArmScores,
     CONTROL_MIN_COMPOSITE_GAP, ControlVerdict, EFFECT_NOISE_RATIO, EffectVerdict, EvalArm,
-    control_sampling, replicate_seeds,
+    PairedEffect, control_sampling, replicate_seed_set, replicate_seeds,
 };
 use gglib_core::domain::benchmark::tune::result::TuneTaskResult;
 use gglib_core::domain::benchmark::tune::task::{ExpectedOutcome, TuneTask};
@@ -203,7 +203,13 @@ pub async fn run_agentic_eval(
     // control's scores to the pipeline.
     let raw_results = take_arm(&mut arm_results, EvalArm::Raw);
     let gglib_results = take_arm(&mut arm_results, EvalArm::Gglib);
-    let replicate_results = take_arm(&mut arm_results, EvalArm::RawReplicate);
+    // Every A/A pair, in plan order — take_arm removes the first match, so
+    // draining repeatedly yields them in the order they ran.
+    let mut replicate_runs: Vec<Vec<Vec<TuneTaskResult>>> = Vec::new();
+    while let Some(run) = take_arm(&mut arm_results, EvalArm::RawReplicate) {
+        replicate_runs.push(run);
+    }
+    let replicate_results = replicate_runs.first().cloned();
     let control_results = take_arm(&mut arm_results, EvalArm::Control);
 
     // Each arm is scored against *its own* seed count, not the run's: the
@@ -222,6 +228,10 @@ pub async fn run_agentic_eval(
     let raw = score_arm(&raw_results, EvalArm::Raw).unwrap_or_else(|| empty_scores(&config));
     let gglib = score_arm(&gglib_results, EvalArm::Gglib).unwrap_or_else(|| empty_scores(&config));
     let raw_replicate = score_arm(&replicate_results, EvalArm::RawReplicate);
+    let raw_replicates: Vec<_> = replicate_runs
+        .iter()
+        .filter_map(|run| score_arm(&Some(run.clone()), EvalArm::RawReplicate))
+        .collect();
     let control = score_arm(&control_results, EvalArm::Control);
     let delta = AgenticEvalReport::delta_of(&raw, &gglib);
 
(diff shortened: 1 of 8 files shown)


mmogr:
**Live hardware verification — passed.** `gglib benchmark agentic -m Qwen3.5-4B --seeds 11 22 --replicate-pairs 2`:

- Two A/A arms ran on derived disjoint seed sets; the report stores `raw_replicates` (composites 0.580, 0.665) and `replicate_seed_sets` with pair 1 equal to the legacy derivation.
- Drift printed as *"0.095 … (mean over 3 pairwise gaps)"* — hand-checked: |0.723−0.580|, |0.723−0.665|, |0.580−0.665| → mean 0.0953. The multi-pair arithmetic is correct on the wire.
- Effect verdict correctly refused: −0.013 at 0.1× the drift, within noise.
- Paired line rendered with the honest Wilcoxon refusal: `1W–2L–15T over 18 (task, seed) pairs … too few non-tied pairs for a p`; the `paired` record is stored on the report with `p_value: null`.
- Control moved (0.488 gap) with its one-seed annotation; the unmeasured-run guard flagged 4/9 control runs that never reached the model and demoted those means to "a floor, not a measurement".

One copy bug found and fixed on this branch (`fcb49aca`): the degrees-of-freedom caveat was static single-pair wording, so it misstated the df directly beneath a three-gap drift line. It now sizes itself to the verdict's pair count.

Issue #37, opened 2025-12-25
Devstral model fails with HTTP 500: system role not supported


mmogr:
## Problem

The devstral model (mistralai_Devstral-Small-2-24B-Instruct-2512) fails when used in the chat interface with the following error:

```
HTTP 500: {"error":"llama-server returned 500 Internal Server Error: {\"error\":{\"code\":500,\"message\":\"Only user, assistant and tool roles are supported, got system. at row 119, column 111...
```

## Steps to Reproduce

1. Add the devstral model to gglib
2. Start a chat conversation with devstral
3. Send a message that includes a system prompt
4. Observe HTTP 500 error

## Expected Behavior

The model should handle chat messages gracefully without errors.

## Actual Behavior

The model rejects system role messages in the conversation loop, causing the chat to fail completely.

## Impact

- Devstral model is completely unusable in the chat interface
- No workaround available for users
- Other Mistral-style models may have similar issues

PR #783, opened 2026-08-11
feat(benchmark): tune --apply — the drift-gated winner write

mmogr:
> **Stacked on #782** (→ #780 → #779). Base is `feat(benchmark)/eval-gates`; retarget as the stack merges.

Fourth PR of the closed-loop arc: the loop's write half, with a human still pressing the button. A tune run's winner can now become a model's `Measured` defaults — but only through a gate whose every rule is the codified form of a failure this repo has already measured.

## The problem this replaces

Both existing apply paths were ungated. The CLI's `--apply-best` and the GUI's "apply best config" checkbox wrote the top composite straight into the model's defaults — no drift check, no incumbent comparison, and without touching `defaults_origin` (leaving an auto-detected model's origin claiming a recipe that no longer matched it). That is precisely the noise-ratchet an auto-tuner becomes without a gate: it stores whichever candidate a lucky draw favoured and reports improvement while doing it.

## The in-run calibration pair

A tune run has no A/A arm, so the drift estimate is built into the candidate list: every run now carries the **incumbent** twice. The incumbent is an all-`None` overlay — it resolves through the normal chain, so it is *exactly what the model does today* — and the gap between the identical twins is the run's own noise, measured under the same tasks, server, and everything else. Both twins always run the full suite, exempt from pruning (a pruned twin leaves the run uncalibrated, and a pre-screen composite is not comparable with full-suite scores). Two extra candidates is the price of a run that can calibrate itself.

## The gate

`evaluate_apply` refuses with a first-class verdict naming its evidence — never an error:

| verdict | rule | the failure it codifies |
|---|---|---|
| `Apply` | margin ≥ `EFFECT_NOISE_RATIO` × the twins' gap, pairs agree | — |
| `IncumbentStands` | no candidate beat the current defaults | a successful run whose answer is "change nothing" |
| `WithinDrift` | margin inside the run's own noise | the +0.082 that did not replicate |
| `PairedDisagrees` | the winner's mean rests on a minority of per-task pairs | the lucky-outlier shape |
| `Uncalibrated` | no incumbent pair (any pre-existing run) | a comparison in which nothing could vary |
| `Contaminated` | unmeasured task runs in the compared candidates | the arm of 45 zeros that rendered as a score |

The direction check reuses PR 3's `PairedEffect` (new `from_paired_runs`, pairing two candidates' task lists by `task_id`) rather than reinventing paired comparison.

## The write, and its trail

An `Apply` verdict stores the winner's **sparse overlay** as `inference_defaults` (unswept dimensions stay `None` and keep resolving through the chain, exactly as they did during the run) with `defaults_origin = Measured` — below global, ceiling-exempt, per #780. The run row records an `ApplyRecord` (verdict numbers + the applied overlay) in a new nullable `applied_json` column — the metadata trail #780 deferred to the apply's own schema — written *after* the model, deliberately: a record naming an apply that never happened misleads, while an apply missing its record is recoverable from the model's origin alone.

## Surfaces

- **HTTP**: `POST /api/benchmark/tune/{run_id}/apply` → `ApplyOutcome` JSON. Always 200 for a verdict; errors are reserved for a missing run, a non-tune run, or storage failing.
- **CLI**: `--apply` (alias `--apply-best` kept for scripts) captures the run id from `RunComplete` and renders every verdict with its numbers.
- **GUI**: the "apply best" checkbox becomes the gated path with verdict messages; the incumbent rows are labelled on the leaderboard (`incumbent` / `incumbent (A/A)`). The **per-row apply deliberately stays ungated**: a person picking one exact candidate is a deliberate choice and writes as user-set defaults, the same as typing the values into `gglib model update` — the gate governs the automatic path, not the person. The distinction is documented on both handlers.

## Verification

- Full workspace `cargo test` green (1033 in `gglib-core`; 7 new gate tests covering every verdict, including a pruned candidate never winning). Clippy `-D warnings` clean, fmt clean, `check_boundaries.sh` passes.
- TS: 816 tests green, `tsc --noEmit` clean, eslint clean.
- Schema: one nullable `ALTER TABLE benchmark_runs ADD COLUMN applied_json` following the repo's ignore-if-exists migration pattern; `BenchmarkRun` deserializes legacy rows via serde default.

### Not verified here

A live gated apply end-to-end (needs a model): `gglib benchmark tune -m <model> --sweep temperature=0.7,1.0 --apply` should show the two incumbent rows in the run, then either an apply with margin-vs-drift numbers or a refusal with evidence, and `gglib model explain` should afterwards say `measured by a tune sweep` on an applied model.

## What this sets up

PR 5 puts this exact call on the daemon's idle-time schedule — the system-pressed button. Nothing about the gate changes there; that is the point of landing it here first.

### DIFF
--- a/crates/gglib-app-services/src/benchmark/mod.rs
+++ b/crates/gglib-app-services/src/benchmark/mod.rs
@@ -149,4 +149,12 @@ impl BenchmarkOps {
     ) -> Result<()> {
         agentic::run_agentic_eval(&self.deps, config, tx, cancel).await
     }
+
+    /// Evaluate a completed tune run against the apply gate and, when the
+    /// verdict licenses it, store the winner as the model's `Measured`
+    /// defaults. Refusals return as verdicts, never as errors — see
+    /// `tune::apply_run`.
+    pub async fn apply_tune_run(&self, run_id: i64) -> Result<tune::apply_run::ApplyOutcome> {
+        tune::apply_run::apply_tune_run(&self.deps, run_id).await
+    }
 }
--- a/crates/gglib-app-services/src/benchmark/mod.rs
+++ b/crates/gglib-app-services/src/benchmark/mod.rs
@@ -18,7 +18,7 @@ mod compare;
 pub mod guard;
 pub mod mapper;
 mod perf;
-mod tune;
+pub mod tune;
 
 // ────────────────────────────────────────────────────────────────────────────
 // Dependency bundle
--- a/crates/gglib-app-services/src/benchmark/tune/apply_run.rs
+++ b/crates/gglib-app-services/src/benchmark/tune/apply_run.rs
@@ -1,0 +1,104 @@
+//! Applying a tune run's winner as a model's `Measured` defaults, through
+//! the gate.
+//!
+//! This is the write half of the closed loop's first cycle: the gate
+//! (`gglib_core::domain::benchmark::tune::apply`) judges the stored run, and
+//! only an [`ApplyVerdict::Apply`] licenses touching the model. Every other
+//! verdict is returned to the caller as itself — a refusal that names its
+//! evidence, never an error.
+
+use anyhow::{Context as _, Result, anyhow};
+use gglib_core::domain::DefaultsOrigin;
+use gglib_core::domain::benchmark::BenchmarkRunType;
+use gglib_core::domain::benchmark::tune::apply::{
+    ApplyRecord, ApplyVerdict, evaluate_apply, winning_candidate,
+};
+use serde::{Deserialize, Serialize};
+
+use super::super::BenchmarkDeps;
+
+/// What an apply attempt returned to its caller.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ApplyOutcome {
+    /// The gate's decision, with its numbers.
+    pub verdict: ApplyVerdict,
+    /// The model the run tuned.
+    pub model_id: i64,
+    /// Whether the model was actually written — `verdict.applies()`, echoed
+    /// so a wire consumer does not have to know the enum's variants to ask
+    /// the one question that matters.
+    pub applied: bool,
+}
+
+/// Evaluate a completed tune run against the apply gate and, if it passes,
+/// store the winner as the model's `Measured` defaults.
+///
+/// The write is the whole of what changes: `inference_defaults` becomes the
+/// winner's sparse overlay (unswept dimensions stay `None` and keep
+/// resolving through the chain, exactly as they did during the run), and
+/// `defaults_origin` becomes [`DefaultsOrigin::Measured`] — below global
+/// settings, exempt from the agentic ceiling, per its own docs. The run row
+/// records the [`ApplyRecord`] so the model's provenance stays traceable to
+/// the numbers that licensed it.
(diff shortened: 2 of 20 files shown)


mmogr:
**Live hardware verification — passed, and the loop closed for the first time.** `gglib benchmark tune -m Qwen3.5-4B --sweep temperature=0.7,1.0 --apply`:

- The run carried 5 candidates: 2 grid points, the qwen-coding family preset, and the incumbent calibration pair — both twins ran the full suite unpruned.
- The gate applied: winner 0.728 (the family preset) over incumbent mean 0.617, margin +0.111 against drift 0.000, paired 0W–0L–9T (no disagreement).
- The write landed end to end: `defaults_origin = measured` in the DB, the winner's sparse overlay as `inference_defaults`, the full `ApplyRecord` (verdict numbers + applied config) in the new `applied_json` column on run row #38.
- `gglib model explain` renders *"per-model defaults (measured by a tune sweep)"* per field, with the coupled trio resolving correctly around the measured layer; `model inspect` shows the "(measured by a tune sweep — ranks below global settings)" suffix. (#780's surfaces, verified in the same pass.)

**One observation for PR 5's design, not a defect here:** the incumbent twins scored identically, so drift = 0.000 — and at zero drift the ratio gate passes any positive margin, leaving the paired check as the only guard. Tune tasks run unseeded and the 9-task suite scores coarsely, so identical twins will not be rare. PR 5 should consider a small drift floor or per-run seed stamping for tune candidates before the automated path leans on this gate.

mmogr:
**Decision on the drift=0 degeneracy (per project lead's delegation): tune tasks are now seeded — no invented drift floor.** `1431201d` + `09148c62` on this branch.

Design mirrors the agentic eval's primary-vs-A/A split: every candidate runs task T under the same derived seed (common random numbers — candidate comparisons stay tightly paired), and the calibration twin alone runs offset seeds (`REPLICATE_SEED_OFFSET`, the A/A arm's constant and rationale). The twins' gap then genuinely samples seed-to-seed variance under the incumbent's own configuration — exactly the noise the gate's question is about. Seeds derive from the task *id* (djb2), so they survive suite reordering and a surprising score can be re-run rather than chased. The gate itself is unchanged: the fix makes the drift a measurement, per the arc's rule, rather than guessing a threshold.

Live re-verification with seeded twins is running now; result to follow.

mmogr:
**Seeded calibration re-verified live.** Run #39 (branch build, worktree DB): the incumbent twins scored 0.617 vs 0.654 — **drift 0.037, a real reading** where the unseeded run measured 0.000 — and the gate refused with `PairedDisagrees (0W–1L)`: the winner's mean cleared the drift bar but rested on an outlier while losing the per-task pairs. Notably the incumbent this time *was* the measured recipe applied by run #38, so the loop judged fresh candidates against its own previous output and declined to churn — the anti-ratchet working end to end.

Two operational notes from the session, both now PR 5 design items: (1) kill any auto-spawned daemon after rebuilding, or the old binary serves the new CLI (bit twice); (2) `ensure_daemon` happily used an older *installed* daemon lacking this PR's apply route — an opaque 405 where a build-fingerprint check would have said 'daemon is older than this CLI, restart it'. One additive tune-run row (#37) landed in the main catalog DB from that mix-up; model defaults there were never touched.

PR #794, opened 2026-08-12
feat(benchmark): the repair-rate signal — a temperature sweep for malformed tool calls

mmogr:
## What

The auto-tune scheduler's second production signal: a per-model **tool-call repair attempt rate ≥ 5% over ≥ 50 windowed requests** now earns a signal-driven **temperature sweep** (`0.2 / 0.5 / 0.8`), stamped `signal:repair-rate` on the run. PR 6 (#786) left the repair counters recorded but unacted, reserved for PR 7's grammar learning — which was permanently parked on the endpoint blocker (#787, ADR 0005). A sampling response to the same failure shape doesn't touch that blocker at all.

## Design decisions

- **Attempt rate, not success rate**: the defect being diagnosed is the malformed call, not whether `tool_choice: "required"` then fixed it. The measured rate **undercounts** on busy traffic (attempts are streaming-only, and the ledger write is skipped when the request's snapshot has left the 50-entry metrics ring) — a fail-safe bias: the signal fires late, never spuriously. Documented on the threshold const.
- **Cross-kind ranking by severity**: with two signal kinds, models compete on *rate ÷ threshold* rather than raw percentages, each model contributing its own worst signal via the new pure `worst_signal()`. A dead heat prefers the loop guard — deterministically, inside the function, because `max_by` keeps the *last* maximum and push order must not be the tie-break.
- **Temperature candidates are whole recipes**: naming `temperature` claims the coupled set under `resolve_layers`, so companions resolve to llama.cpp neutral defaults — exactly how an applied temperature-only Measured recipe resolves in production. Measurement equals shipped behaviour. Swept values are never ceiling-clamped (trusted top layer in the eval; Measured model rung ceiling-exempt post-apply, #748). Documented on `SignalKind::sweep()`.
- **Bypass contract unchanged**: signals still bypass exactly two rules (untuned-only filter, retune interval) and honour the rest — ADR 0005's sentence updated to name both signals.

## Surfaces

- Frontend activity card and CLI `run_initiator` are slug-generic — `signal (repair-rate)` / `auto (signal:repair-rate)` render with no code changes; the CLI Initiator column widens 22 → 25 to fit the new slug.
- Docs: ADR 0005 (signal sentence + a dated note in the lazy-grammar section), `docs/tool-call-repair.md` Observability, `crates/gglib-proxy/README.md` counter table.

## Tests

Six new behaviour-named tests beside the existing ten (all green unchanged): loud repair rate fires the temperature sweep; thin sample and quiet rate never signal; severity ranking across kinds; one model's worst defect names the sweep; dead heat prefers the loop guard. Full `make pre-commit` green.

## Live verification

Plumbing verified via the unit suite; the end-to-end signal fire (seeded window → temperature sweep → gated apply) is exercised by the smoke test in the stacked persistence PR, which makes seeding a repair window trivial.

### DIFF
--- a/crates/gglib-app-services/src/benchmark/auto_tune.rs
+++ b/crates/gglib-app-services/src/benchmark/auto_tune.rs
@@ -22,9 +22,11 @@
 //!   fully cold GPU tunes from the catalog at large.
 //! - **A person's work is never the target.** User-set defaults are
 //!   excluded outright — the same principle that ranks `Measured` below
-//!   global settings — and measured models are excluded because re-checks
-//!   are a deliberately deferred decision (PR 6's signal triggers), not a
-//!   timer's.
+//!   global settings — and measured models are excluded from the idle
+//!   queue because re-checks belong to the signal triggers: a production
+//!   defect rate past its threshold (a loop-trip rate → a DRY sweep, a
+//!   tool-call repair rate → a temperature sweep) is the one thing that
+//!   reopens a measured recipe, not a timer.
 //! - **Any real request preempts.** While a run is in flight the scheduler
 //!   watches the admission queue; the moment anything is waiting, the run
 //!   is cancelled and the GPU handed over. The next attempt starts from a
@@ -85,7 +87,7 @@ const PREEMPT_POLL: Duration = Duration::from_secs(2);
 
 /// The least time between tune attempts on one model, however they ended.
 /// A refusal is an answer; the next question should wait for new evidence
-/// (a different build, new tasks) or PR 6's signal triggers.
+/// (a different build, new tasks) or the signal triggers.
 const RETUNE_INTERVAL: chrono::Duration = chrono::Duration::days(7);
 
 /// The fewest windowed requests before a defect rate means anything.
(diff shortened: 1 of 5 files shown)


mmogr:
Closing unmerged, but **harvesting** — this one contains work we want, attached to a mechanism we're removing.

The PR adds a repair-rate signal that triggers a temperature sweep from the idle-time scheduler. The scheduler is being deleted in the adaptive-inference transition, so the trigger half has nowhere to live. The *measurement* half is still exactly right.

Being carried forward into `gglib-core`:

- `SignalKind` and its `Display`
- `MIN_SIGNAL_SAMPLE` and the three thresholds
- `sweep()`
- `worst_signal()`

Deliberately **not** carried forward: the initiator slug, and both Initiator-column width commits — the column collapses to a constant once nothing but manual runs exist, and dropping it reclaims 28 columns for the narrow-terminal work.

One sequencing note for whoever picks this up: harvesting these into `main` early would leave them unreferenced for several branches, since the code that consumes them doesn't exist yet. They land with the scheduler-removal branch, or with the telemetry that reads them — not before.

The branch is left in place so the harvest can be done from it directly.

PR #795, opened 2026-08-12
feat(ledger): defect-window persistence — evidence that survives restarts, decayed and build-scoped

mmogr:
## What

Stacked on #794. The defect ledger's evidence no longer dies with the process: the auto-tune scheduler persists its per-model **unacted windows** every tick and restores them at boot — decayed by wall-clock age and discarded across a llama.cpp release change. On a restart-heavy host, the ≥ 50-request sample floor can now actually be reached; previously the signals quietly starved.

## Design

- **The window, not the ledger.** What lands in `defect_windows` is `delta(ledger, baselines)` — evidence not yet spent on a sweep. Spent evidence is subtracted *before* persist, so boot seeds the ledger with the rows as-is, leaves baselines empty, and double-fire is structurally impossible rather than carefully avoided.
- **Two answers to ADR 0005's staleness objection** (the paragraph is retracted in place, struck and dated per the house rule): exponential decay with a **7-day half-life** (`GGLIB_DEFECT_HALF_LIFE_DAYS` to override; rates are decay-invariant, so the half-life really governs how long old traffic counts toward the sample floor), and **build scoping** via `effective_llama_release()` — single-sourced from the installer's own selector policy — so another build's rate is deleted, not decayed.
- **Two crash/consistency edges closed deliberately:**
  - the spent window is zeroed in the store *before* a signal-driven run starts — a crash mid-run cannot leave the loud pre-signal row behind to re-seed and re-fire;
  - unchanged rows are never rewritten, which preserves a stored row's `updated_at` so untouched evidence keeps decaying from its real age across consecutive restarts.
- **Flush sits above the `auto_tune` setting check** — evidence accumulates (and survives restarts) while the feature is off, so enabling it later starts from real history. A final flush rides the scheduler's own cancellation; `perform_shutdown` needs no hook and the write is a single local upsert, far inside the 10s watchdog.
- **No new Settings field**; the half-life is a const with the established env-override idiom. `ModelDefectLedger::new()` keeps its signature (seeding is a new additive `seed()`), so all ten existing construction sites are untouched.

## Layers

domain (`PersistedDefectWindow`, `seed`) → port (`DefectWindowRepositoryPort`) → schema + `SqliteDefectWindowRepository` → `effective_llama_release()` (gglib-runtime, prebuilt-gated) → wiring (one adapter call site, beside `bench_repo`) → scheduler (`SchedulerState`, restore/flush, pure decay fns) → ADR 0005 retraction + README index entry (which was missing ADR 0005 entirely).

## Tests

16 new: decay boundary cases (zero gap, one half-life exact, huge gap, future timestamp), restore-plan policy (foreign build discarded, decayed-below-one-request discarded, seeds decayed), flush policy (unchanged never rewritten, changed row carries build + stamp), the end-to-end restore contract (persisted loud window → fresh ledger → empty baselines → repair-rate signal fires), ledger seeding (windows through the normal delta path, saturation), sqlite round-trip/upsert-replace/targeted-delete, and the release-scope string. Full `make pre-commit` green.

## Live smoke (documented, not yet run — the installed daemon holds port 9887 mid-download)

```bash
export GGLIB_DATA_DIR=$(mktemp -d)
GGLIB_AUTO_TUNE_TICK_SECS=2 GGLIB_AUTO_TUNE_IDLE_TICKS=2 target/debug/gglib daemon run &
# import a small model; gglib config auto-tune on; stop the daemon
sqlite3 "$GGLIB_DATA_DIR/gglib.db" ".schema defect_windows"
# 1. seed a loud repair window against the current pin:
sqlite3 "$GGLIB_DATA_DIR/gglib.db" "INSERT INTO defect_windows VALUES ('<model>', 100, 0, 12, 6, strftime('%Y-%m-%d %H:%M:%S','now'), 'b10327');"
# 2. restart → boot log shows the restored window; after 2 idle ticks the repair-rate
#    signal fires a temperature sweep (initiator signal:repair-rate in benchmark list + GUI)
# 3. post-run the row has collapsed to zero; restart again → no refire
# 4. updated_at 60 days back → discarded at boot;  5. llama_build 'b0001' → discarded
```
(Kill the daemon after every rebuild — the stale-binary trap has hit four times in this arc.)


mmogr:
Closing unmerged — the decision went the other way, deliberately.

This PR persists the scheduler's unacted defect windows so a daemon restart resumes the count instead of zeroing it, with exponential decay by wall-clock age and outright discard of evidence recorded against a different llama.cpp release.

I re-authored the pure half onto `main` (#804), then removed it again (#805) once the reasoning was followed through. The short version:

**Decay and build scoping were never features in their own right.** They were the price of persisting at all — invented to answer the staleness objection that `domain/defects.rs` had already raised against persistence in its own module doc:

> *"Deliberately not persisted: a defect rate is a claim about recent traffic on this build of everything, and yesterday's rate answering today's question is exactly the staleness ADR 0001 warns about."*

That was the considered position. This PR retracted it in order to serve the idle-time tune scheduler. With the scheduler being removed in the adaptive-inference transition, the retraction has nothing left to stand on.

**And nothing else was reading it.** Grepping for consumers of `PersistedDefectWindow`, `DefectWindowRepositoryPort` or the restore path outside the module itself returned nothing.

The strategic reason outweighs the tidiness one: sampling defaults now come from the model's own GGUF `general.sampling.*` metadata and its publisher's `generation_config.json` — a claim about the model, not a claim about last week's traffic. Measured per-model defect rates were only ever an input to the tuner being deleted.

What survives from this line of work: `stream_errors` (via #803) is now a live per-model counter, and the counters remain useful as diagnostics — just per-process, resetting on restart, which is the right lifetime for a signal whose whole claim is "what is happening now".

Everything here is recoverable from this branch and #804 if that judgement ever needs revisiting. `domain/defects.rs` now records that persistence was built and removed, and why, so it doesn't get rebuilt by someone reading the transition plan cold.

PR #797, opened 2026-08-12
fix(benchmark): an aborted tune run no longer consumes the retune interval

mmogr:
## What

`select_target`'s interval rule counted **any** tune run in the last seven days, whatever its status. An aborted run — preempted, crashed at launch, or killed by a daemon stop mid-flight — therefore parked its model for a week having produced no information at all.

Only an *answered* question holds a claim now: `Complete` (its verdict stands for the interval) and `Running` (its answer is coming, and two surfaces must not race the same question). `Failed` holds nothing.

## Why it matters, live

Both models on the author's machine are currently parked by aborted runs:

```
1|tune|failed|[1]|2026-08-12 04:25:14|Aborted by user
2|tune|failed|[2]|2026-08-12 07:05:01|Process terminated unexpectedly
```

Run 1 was the first-ever unattended idle tune, killed seconds in by a daemon stop. Run 2 was a recovery tune — the scheduler measuring a deliberately sabotaged recipe — killed by a daemon restart. With the daemon now fully idle and eager (44 idle ticks banked), the log reads `auto-tune: idle, but nothing eligible to tune` on every tick: the scheduler is standing down for a week on two runs that measured nothing.

## The trade, accepted and documented

A tune that crashes *deterministically* now retries each idle window rather than backing off for a week. Bounded by the idle threshold (10 sustained idle minutes per attempt) and visible in the activity feed with its initiator, so a repeating failure is loud rather than silent. The alternative — treating a crash as an answer — is what this PR removes.

## Tests

`an_aborted_run_does_not_park_the_model` and `a_running_tune_parks_the_model`, beside the existing interval test. Full `make pre-commit` green.

Stacked on #795.

### DIFF
--- a/crates/gglib-app-services/src/benchmark/auto_tune.rs
+++ b/crates/gglib-app-services/src/benchmark/auto_tune.rs
@@ -92,9 +92,16 @@ fn env_override<T: std::str::FromStr + Copy>(key: &str, default: T) -> T {
 /// How often the preemption watcher checks the queue while a run is live.
 const PREEMPT_POLL: Duration = Duration::from_secs(2);
 
-/// The least time between tune attempts on one model, however they ended.
+/// The least time between *answered* tune attempts on one model.
 /// A refusal is an answer; the next question should wait for new evidence
-/// (a different build, new tasks) or the signal triggers.
+/// (a different build, new tasks) or the signal triggers. An aborted or
+/// crashed run is **not** an answer — a preempted idle tune, a daemon
+/// stopped mid-run, a launch failure — and must not park the model for a
+/// week (the first live idle tune was aborted by a daemon stop and parked
+/// its model seven days for zero information). Only completed runs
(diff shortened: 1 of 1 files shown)


mmogr:
Closing unmerged — not because the fix is wrong, but because its subject is being deleted.

This corrects `RETUNE_INTERVAL` accounting inside the idle-time auto-tune scheduler. That scheduler is being removed entirely in the adaptive-inference transition: the tuner kept rediscovering the published preset, a 27B tune costs 2–3 hours, and the models most needing correction are the most expensive to correct. Merging this would land a fix to code that the scheduler-removal branch then deletes wholesale.

The reasoning is worth keeping even though the code isn't — **"an aborted run answered nothing and holds no claim"** is the right principle for any future cooldown or backoff, and it's recorded in the transition plan for exactly that reason.

The branch is left in place; nothing here is lost if we want it back.

PR #798, opened 2026-08-12
feat(proxy,benchmark): the stream-error signal — the counter that sees dead, not just sick

mmogr:
## What

A third defect counter and its signal: `stream_errors` counts turns that died on an **upstream mid-stream failure** — the model server emitting an error event mid-generation, or the byte stream itself breaking. At ≥ 5% over ≥ 50 windowed requests it raises `signal:stream-error`, treated with the same temperature sweep as the repair-rate signal.

## Why — a blind spot found by sabotage, not by argument

A model's recipe was deliberately sabotaged (temperature 2.0, top_p 1.0, top_k off) to make the existing signals fire on organic traffic. They never did. The reason is structural: **both existing counters require a model coherent enough to produce structured output.** The loop guard counts verbatim repetition; the repair loop counts a tool call that was *attempted and malformed*. A model whose sampling has collapsed produces neither — it emits output so far outside the expected shape that the model server kills the stream.

A real editor session died exactly that way, with nothing recorded:

```
Reason: Chat completion request failed: Inference server reported an error
mid-stream: The model produced output that does not match the expected
peg-native format
```

| Failure mode | Counted before | Counted now |
|---|---|---|
| Verbatim loops | ✅ loop-guard trips | ✅ |
| Malformed-but-complete tool calls | ✅ repair attempts | ✅ |
| Output so broken the stream is killed | ❌ **invisible** | ✅ stream errors |

The counters could see *sick*. They could not see *dead* — the failure a person actually feels, because their chat simply fails.

## Correctness details

- **Client disconnects never count.** Hanging up is a person's action, not a model defect. Only the two upstream-failure sites in the drain loop set the flag: an `UpstreamError` event (split out of the `ToolCallDelta` arm it shared) and the byte-stream `Err` arm.
- **The counter never bumps its own denominator.** Unlike a loop-guard trip (which fires *instead of* a forward), a stream error marks a request already counted at forward time. Test: `a_stream_error_never_bumps_its_own_denominator`.
- **Same ring-eviction undercount** as tool repairs, and the same fail-safe direction: the signal fires late, never spuriously.
- **Three-way severity** with ties resolved in array order — loop guard, then stream error, then repair — inside the fold, so ordering never depends on `max_by` keeping the last maximum.
- **The migration runs after its CREATE**, per the #796 lesson; existing `defect_windows` rows backfill to 0.

## Tests

`a_dying_stream_earns_a_temperature_sweep`, `a_dead_heat_prefers_death_over_repair`, `a_stream_error_never_bumps_its_own_denominator`, plus the persistence round-trip and decay fixtures extended with the new field. Full `make pre-commit` green.

Stacked on #797.

### DIFF
--- a/crates/gglib-app-services/src/benchmark/auto_tune.rs
+++ b/crates/gglib-app-services/src/benchmark/auto_tune.rs
@@ -24,9 +24,9 @@
 //!   excluded outright — the same principle that ranks `Measured` below
 //!   global settings — and measured models are excluded from the idle
 //!   queue because re-checks belong to the signal triggers: a production
-//!   defect rate past its threshold (a loop-trip rate → a DRY sweep, a
-//!   tool-call repair rate → a temperature sweep) is the one thing that
-//!   reopens a measured recipe, not a timer.
+//!   defect rate past its threshold (a loop-trip rate → a DRY sweep; a
+//!   tool-call repair rate or a mid-stream failure rate → a temperature
+//!   sweep) is the one thing that reopens a measured recipe, not a timer.
 //! - **Any real request preempts.** While a run is in flight the scheduler
 //!   watches the admission queue; the moment anything is waiting, the run
 //!   is cancelled and the GPU handed over. The next attempt starts from a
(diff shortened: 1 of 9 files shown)


mmogr:
Superseded by #803, which is now merged. Closing.

#803 re-authors the durable half of this PR onto clean `main`. What carried over unchanged is the idea and most of the code: `stream_errors` as a per-model counter, flagged from the two drain-loop sites that already know a turn died, back-patched through the metrics ring like the repair flag.

Three things were deliberately left behind:

- **The `setup.rs` `ALTER TABLE defect_windows ADD COLUMN stream_errors`.** It hard-depends on #795 creating that table, so it folds into #795 instead — the column gets created correctly the first time rather than added by a migration that has to run in the right order.
- **The `auto_tune.rs` (+116) and `benchmark.rs` changes.** Scheduler wiring for a scheduler being removed in the adaptive-inference transition.
- **The ADR 0005 edits.** ADR 0005 is being superseded by ADR 0006 rather than amended — it already carries one strikethrough retraction, and stacking a second would make it unreadable as a record.

Also worth recording: this PR described its work as "the proxy half", but the ledger it extends lives in `gglib-core/src/domain/defects.rs`, not `gglib-proxy`. #803 is scoped `(proxy,core)` accordingly.

Two of the facts this PR needed had already landed separately in #800 — `StreamOutcome::upstream_errored` and the `UpstreamError`/`ToolCallDelta` match-arm split — which is why #803 is so much smaller than this.

PR #843, opened 2026-08-14
fix(ci): the settings-surface guard actually checks the CLI again

mmogr:
## What

`scripts/check_settings_surfaces.sh` has been running at half strength since
333682f3 split `config_commands.rs` into a directory. The guard kept pointing at
the old path, so its CLI arm was dead: `grep` printed *No such file or
directory* once per field and, because the call sits inside an `if`,
`set -euo pipefail` never aborted. `in_cli` was always false and every setting
passed on the GUI check alone.

Before this change, on `main`:

```
grep: crates/gglib-cli/src/config_commands.rs: No such file or directory
  ✓ tool_call_repair (GUI)
  ...
✅ every Settings field is reachable (23 checked)      # exit 0
```

After:

```
  ✓ tool_call_repair (CLI, GUI)
  ...
✅ every Settings field is reachable (23 checked, 18 on the CLI)
```

18 matches the number of `Option<…>` flags in `SettingsSetArgs` exactly.

## Why two changes, not one

Repointing the path alone would have kept matching nothing. The fields moved
from a clap variant's inline list to a named `Args` struct, so they now carry a
`pub ` prefix the old pattern did not allow for. Both halves are needed.

## Why the second commit

The bug was never really the wrong path — it was that a wrong path produced
output indistinguishable from a healthy run. The second commit adds one rail per
way the CLI arm can stop working:

| Failure | Rail |
|---|---|
| File missing | Checked up front, instead of `grep` failing inside an `if` where `set -e` cannot see it |
| Pattern matches everything | Probes a name that is not a field; refuses to continue if it matches |
| Pattern matches nothing | Several settings are deliberately CLI-only, so zero CLI-reachable fields is not a real state — it fails instead of reporting all clear |

Each was verified by perturbing a copy of the script and confirming the intended
rail fires:

```
variant 1 (old path)              exit=1  ❌ crates/gglib-cli/src/config_commands.rs does not exist
variant 2 (pattern too loose)     exit=1  ❌ self-test: the CLI detector matches a field that does not exist
variant 3 (pattern too strict)    exit=1  ❌ no Settings field was found on the CLI surface
```

This mirrors `check_param_source_exhaustive.sh`, which already refuses to trust
its own verdict when it cannot prove it still detects.

## Context

Found during a dead-code audit of the workspace. This guard is the one written
because `tool_call_repair` sat unreachable for months, so it seemed worth fixing
before anything else in that sweep.

## Verification

- `./scripts/check_settings_surfaces.sh` — 23 checked, 18 on the CLI, exit 0
- `make enforce` — passes
- No Rust or TypeScript is touched

### DIFF
--- a/scripts/check_settings_surfaces.sh
+++ b/scripts/check_settings_surfaces.sh
@@ -44,6 +44,17 @@ cd "$ROOT_DIR"
 SETTINGS_RS="crates/gglib-core/src/settings.rs"
 CLI_SET_ARGS="crates/gglib-cli/src/config_commands/settings_args.rs"
 
+if [ ! -f "$CLI_SET_ARGS" ]; then
+  echo -e "${RED}❌ ${CLI_SET_ARGS} does not exist${NC}"
+  echo
+  echo "The CLI half of this guard reads that file. When the path is wrong the"
+  echo "grep below fails per field, and because it sits inside an \`if\` it does"
+  echo "not trip \`set -e\` — every field then passes on the GUI check alone."
+  echo "That is exactly how this guard ran half-blind after 333682f3 moved"
+  echo "\`config_commands.rs\` into a directory."
+  exit 1
+fi
+
 # Fields nothing settable is expected for, and why.
 declare -A EXEMPT=(
   [setup_completed]="written by the setup wizard when it finishes, not chosen"
@@ -89,8 +100,27 @@ to_camel() {
   }'
 }
 
+# CLI detection, factored into a function so the self-test below exercises the
+# same code path the real scan uses. The `pub ` is optional because the fields
+# moved from a clap variant's inline list (no `pub`) to a named `Args` struct
+# (`pub`) — a pattern that assumes either one silently matches nothing.
+cli_has_field() {
+  grep -qE "^[[:space:]]+(pub )?$1: Option<" "$CLI_SET_ARGS"
+}
+
+# A name that is not a field must not match. This catches the opposite failure
+# from the one above: a pattern loose enough to match everything reports every
+# field as CLI-reachable and is just as useless as one that matches nothing.
+if cli_has_field "definitely_not_a_settings_field"; then
+  echo -e "${RED}❌ self-test: the CLI detector matches a field that does not exist${NC}"
+  echo
+  echo "The pattern is too loose, so every field will look CLI-reachable."
+  exit 1
+fi
+
 unreachable=()
 checked=0
+cli_reachable=0
 
 for field in $fields; do
   checked=$((checked + 1))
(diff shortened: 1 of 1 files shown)


mmogr:
Good catch — the GUI half had the identical bug and I'd only fixed one side of it. Reproduced on the branch as it stood: renaming `src/` gave

```
  ✓ close_to_tray (CLI)
  ✓ start_at_login (CLI)
✅ every Settings field is reachable (23 checked, 18 on the CLI)     # exit 0
```

Pushed 2635eacb.

**GUI symmetry** — done as asked: `[ -d src ]` up front, a factored `gui_has_field`, a self-test that a non-field doesn't match, a `gui_reachable` counter, a zero-assertion, and `2>/dev/null` removed now that the directory is checked.

**Tightened assertion** — implemented, with one change of shape. Comparing `declared` to `cli_reachable` as totals couples two counts that aren't required to be equal: `declared` counts every flag in `SettingsSetArgs`, while `cli_reachable` counts non-exempt `Settings` fields. Those are 18 and 18 today by coincidence, not by construction — give an exempt field a CLI flag, or add a flag that isn't a `Settings` field, and a correct change fails CI with a confusing message. So the cross-check compares *sets*: every flag the struct declares, that `Settings` also declares and doesn't exempt, must be found by the detector. The enumeration uses a deliberately looser pattern than the detector, so drift between them is what fails.

It catches what you were aiming at. Perturbing the detector to `Option<u` — matching numeric fields but not `Option<String>`/`Option<bool>` — leaves six fields detected, so the zero-check waves it through, and the cross-check stops it:

```
❌ the CLI detector missed 12 flag(s) that SettingsSetArgs declares
  - default_download_path
  - show_memory_fit_indicators
  - bind_host
  ... (12 total)
```

**One thing I did not add**, and want to flag rather than quietly skip: the GUI equivalent, `gui_reachable` vs total fields. This guard's rule is that *one* surface is enough — its header calls out `close_to_tray` as deliberately GUI-only and scripted installs as deliberately CLI-only. Requiring every field on the GUI passes today only because all 18 non-exempt fields happen to have both surfaces, and it would fail on the first genuinely CLI-only setting, which the script explicitly permits. That's a policy change rather than a robustness check, so I left it out — happy to add it if you'd rather tighten the policy too, but it should be a deliberate call.

All six rails verified by perturbing a copy of the script:

| Perturbation | Fires |
|---|---|
| Old CLI path | `❌ config_commands.rs does not exist` |
| CLI pattern too loose | `❌ self-test: the CLI detector matches a field that does not exist` |
| CLI pattern too strict | `❌ no Settings field was found on the CLI surface` |
| Wrong GUI path | `❌ src_typo/ does not exist` |
| GUI pattern too loose | `❌ self-test: the GUI detector matches a field that does not exist` |
| CLI pattern partially drifted | `❌ the CLI detector missed 12 flag(s)` |

Green run now reads `23 checked, 18 on the CLI, 18 on the GUI`. `make enforce` passes.

PR #844, opened 2026-08-14
chore(cli,tauri): close the module tree in the two crates nothing imports

mmogr:
First code-touching step of the dead-code sweep. Both crates here are leaves —
`cargo metadata` shows no workspace consumer for either — so everything `pub` in
them is reachable by nobody, which is precisely the state in which `dead_code`
stays silent: the lint has to assume some other crate might be the caller.

## The cascade, measured

The point of demoting the module tree is that it arms the next lint. That is
observable rather than theoretical:

| | `unreachable_pub` in `gglib-cli` |
|---|---|
| Before the root modules were demoted | **20** |
| After | **215** |

Same code, newly auditable. All 215 were then demoted by `cargo fix`, so each is
rustc's own suggestion — it picked `pub(super)` over `pub(crate)` wherever the
only caller is the parent module, which is tighter than I would have written.

Then `dead_code` finally spoke, and found seven things no text search had.

## What the compiler found

| Item | Why it is dead |
|---|---|
| `bootstrap_with` | A test-only composition root no test calls, plus the empty `mod tests` whose only content was a comment saying real tests would use it |
| `run_repl` | A wrapper passing `Vec::new()` to `run_repl_with_prior`; every caller already goes to the latter |
| `format_optional` | A table helper with no caller |
| `DisplayStyle` | `Compact` never constructed → a one-variant enum → held in a `style` field nothing read. Enum and field both go |
| `QueueDownloadDto` | `queue_download` deserialised a queue position and shard count its one caller discards; now checks status and returns `()` |
| `DownloadArgs::force` | Carried three layers, read by nobody |

`gglib-app` is the interesting null result: 132 items were over-public, and
**nothing in it turned out to be dead.** Its visibility was wrong; its contents
are all genuinely used.

## One thing that is not dead code, and is not fixed here

`--force` on `gglib model download` is a real advertised flag with a `-f` short
form and the help text "Skip confirmation prompt". It is inert: `exec` has no
prompt to skip, unlike its `remove` and `update` siblings, which use their
`force` properly.

The dead plumbing comes out here. The flag itself is user-facing surface, so it
is bound and dropped at the destructure with a note rather than being quietly
removed from the CLI inside a visibility change. It belongs with `--weight-speed`
and the other inert flags — I'd suggest folding it into that PR.

## Knock-on effects, stated rather than hidden

- **A doctest becomes `ignore`.** `truncate_string` is now `pub(crate)`, and a
  doctest compiles as its own crate, so no import path can name it. Its two
  assertions already exist verbatim in the module's unit tests. Doctests go from
  3 passing to 2.
- **Docs that named deleted items are updated.** Two doc comments linked
  `run_repl` and would have failed `broken_intra_doc_links`. The presentation
  module README documented `DisplayStyle` and `format_optional`; while fixing it
  I found its `display_model_summary` example was *already* wrong before this PR
   — it passed a `DisplayStyle` to a function taking `ModelSummaryOpts`, naming a
  `Detailed` variant that never existed.
- **Module READMEs no longer use `gglib_cli::` paths** in examples, since those
  modules are crate-internal now.

## Scope

`unreachable_pub` is applied via `RUSTFLAGS` for discovery only — no lint config
is added. Turning it on permanently is the last PR in this arc, once the library
crates have been through the same cascade.

Warnings this surfaced in `gglib-core`, `gglib-hf`, `gglib-runtime` and others
are deliberately untouched; they are the next two PRs.

## Verification

- `make pre-commit` — passes in full
- `cargo clippy -p gglib-cli -p gglib-app --all-targets --all-features -- -D warnings` — clean
- `cargo test -p gglib-cli -p gglib-app` — 244 tests pass
- `cargo test --doc -p gglib-cli` — passes (2, down from 3, per the note above)
- Net: 106 files, 70 lines of dead code removed, the rest visibility

### DIFF
--- a/crates/gglib-cli/Cargo.toml
+++ b/crates/gglib-cli/Cargo.toml
@@ -60,3 +60,15 @@ tokio-test = { workspace = true }
 
 [lints.rust]
 unsafe_code = "deny"
+# Locks in the visibility this crate was just given: every `pub` here must be
+# reachable from outside, and the only things outside are `main.rs` and
+# `tests/`, so in practice it must be in `lib.rs`'s re-export list.
+#
+# Declared per-crate rather than switching this section to `[lints] workspace
+# = true`, which is the usual way to pick up `[workspace.lints]`. Cargo does
+# not let a manifest do both: `workspace = true` must be the only key in
+# `[lints]`, so adopting it here would also switch on the workspace's clippy
+# `pedantic` and `nursery` sets, which this crate has never been held to —
+# 559 warnings, which is its own piece of work rather than a line in a
+# visibility PR.
+unreachable_pub = "deny"
--- a/crates/gglib-cli/src/bootstrap.rs
+++ b/crates/gglib-cli/src/bootstrap.rs
@@ -13,7 +13,7 @@ use anyhow::Result;
 use gglib_bootstrap::{BootstrapConfig, BuiltCore, CoreBootstrap};
 use gglib_core::ports::{
     AppEventEmitter, DownloadManagerPort, GgufParserPort, ModelCatalogPort, ModelRegistrarPort,
-    ModelRepository, NoopEmitter, Repos, SettingsRepository,
+    ModelRepository, NoopEmitter, SettingsRepository,
 };
 use gglib_core::services::AppCore;
 use gglib_db::SqliteBenchmarkRepository;
@@ -154,46 +154,3 @@ pub async fn bootstrap(config: CliConfig) -> Result<CliContext> {
     })
 }
 
-/// Bootstrap with custom repos (for testing).
-///
-/// Note: callers provide their own download manager (typically a stub that
-/// does nothing). This path bypasses [`CoreBootstrap::build`] entirely.
-pub fn bootstrap_with(
-    repos: Repos,
-    downloads: Arc<dyn DownloadManagerPort>,
-    gguf_parser: Arc<dyn GgufParserPort>,
-    model_registrar: Arc<dyn ModelRegistrarPort>,
-    llama_server_path: PathBuf,
-) -> CliContext {
-    let model_repo = repos.models.clone();
-    let app = Arc::new(AppCore::new(repos.clone()));
-    let mcp = Arc::new(McpService::new(
-        repos.mcp_servers.clone(),
-        Arc::new(NoopEmitter),
-    ));
-    CliContext {
-        app,
-        mcp,
-        downloads,
-        gguf_parser,
-        catalog: Arc::new(CatalogPortImpl::new(Arc::clone(&model_repo))),
-        model_repo,
-        model_registrar,
-        llama_server_path,
-        base_port: 9000,
-        http_client: reqwest::Client::new(),
-        bench_repo: Arc::new(SqliteBenchmarkRepository::new_in_memory_blocking()),
-        settings_repo: repos.settings.clone(),
-        download_emitter: Arc::new(CliDownloadEventEmitter::new()),
-    }
-}
-
-#[cfg(test)]
-mod tests {
-
-    #[test]
-    fn test_config_with_defaults() {
-        // with_defaults can fail if paths don't exist, so just test the method exists
-        // In real tests, we'd use bootstrap_with() with mocks
-    }
-}
--- a/crates/gglib-cli/src/bootstrap.rs
+++ b/crates/gglib-cli/src/bootstrap.rs
@@ -153,4 +153,3 @@ pub async fn bootstrap(config: CliConfig) -> Result<CliContext> {
         download_emitter,
     })
 }
-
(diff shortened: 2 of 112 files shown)


mmogr:
All four addressed in 9672a088. Two notes where I did something slightly different from the literal ask, both with evidence.

### 1. `--force` — done as asked

`#[arg(short, long, hide = true)]`, doc comment reworded so it no longer promises a prompt this path doesn't have, and a real `// TODO` marking it for retirement.

Verified it's hidden but still parsed, so nothing that works today breaks:

```
$ gglib model download --force
error: the following required arguments were not provided:
  <MODEL_ID>                       # ← accepted, complains about the missing arg

$ gglib model download --nonsense
error: unexpected argument '--nonsense' found    # ← contrast
```

### 2. Lint inheritance — right catch, and it's worse than one crate

You're correct that `[lints.rust]` doesn't inherit `[workspace.lints]`, so the planned `unreachable_pub = "deny"` would have skipped `gglib-cli`. Checking the whole workspace, **8 of 18 crates are in that position**:

| Shape | Crates |
|---|---|
| Own `[lints.rust]`, inherits nothing | `app-services`, `axum`, `cli`, `db`, `runtime`, `tauri` |
| No `[lints]` section at all | `proxy`, `gglib-app` |

None of them has ever been subject to the workspace's clippy `pedantic`/`nursery` — and that set includes three of the four largest crates in the repo.

**Why I didn't just switch it to `[lints] workspace = true`:** Cargo won't let a manifest do both — `workspace = true` must be the only key in `[lints]` — so adopting it also adopts pedantic and nursery in one step. I measured it on `gglib-cli`:

```
559 warnings
  174  redundant_pub_crate      (nursery)
  109  uninlined_format_args
   77  doc_markdown
   26  struct_field_names
   ...
```

That's its own piece of work, not a line item in a visibility PR — and note the irony in the top entry: `redundant_pub_crate` would want many of the `pub(crate)`s `cargo fix` just added to go fully private. `unreachable_pub` and `redundant_pub_crate` pull in opposite directions, which is worth deciding deliberately rather than discovering mid-merge.

So I declared `unreachable_pub = "deny"` per-crate in both crates this PR cleaned. That achieves the actual goal — the rule now applies where the work was done — and locks it in immediately rather than waiting for the final PR. `gglib-app` got a `[lints.rust]` section for the first time.

Proven to bite, rather than assumed:

```
$ # inject `pub fn deliberately_unreachable_probe() {}` into a private module
crates/gglib-cli/src/utils/input.rs:203:1: error: unreachable `pub` item
src-tauri/src/dock.rs:97:1: error: unreachable `pub` item
```

**Two follow-ups this creates**, which I'd suggest tracking rather than folding in here:
- The final PR of this arc must set `unreachable_pub` per-crate for the other 6 non-inheriting crates, not just at the workspace root — the plan as written would have missed them.
- The pedantic/nursery gap across those 8 crates is a genuine finding independent of this sweep.

### 3 & 4 — done

Comment now says "twelve names" (12 across 8 `pub use` lines). The `truncate_string` example is fenced `text`, so rustdoc renders it as prose rather than an untested example — and as you say, no import path could have made it compile.

### Verification

`make pre-commit` passes. The Rust ratchet caught `model_commands.rs` growing 437 → 440 for the flag's comment; I trimmed the comment first, then recorded the remainder with `--update`, which also ratchets `daemon_client/mod.rs` down 366 → 362 from the earlier deletion. Both visible in the baseline diff.

mmogr:
Heads-up: the `unreachable_pub = "deny"` I added in 9672a088 failed CI, and the reason is worth recording because it will recur in the later PRs of this arc.

`tray/mod.rs` picks its backend by platform:

```rust
#[cfg(not(target_os = "linux"))] mod build;
#[cfg(target_os = "linux")]      mod layer_shell;
#[cfg(target_os = "linux")]      mod linux;
```

So the Linux half was never in the compilation this branch was developed against — `cargo fix` could not have seen it, and neither could my local verification. Six `pub` items came through untouched and became hard errors on Ubuntu the moment the lint was denied. Fixed in 78578113.

**A macOS `cargo fix` is structurally blind to `cfg`-gated code for other platforms.** Nothing in the local toolchain reports this; only CI does.

Each item takes the visibility its own platform twin already has rather than a fresh judgement:

| Item | Now | Matches |
|---|---|---|
| `linux::{Handle, build, sync}` | `pub(super)` | `build.rs`'s three, set by `cargo fix` |
| `layer_shell::{prepare, anchor_at}` | `pub(super)` | called from `placement.rs`, a sibling inside `tray/` |
| `placement::prepare` (Linux) | `pub(crate)` | its `cfg(not(linux))` twin ten lines below |

The reason to expect this to work is that the macOS path already compiles with exactly this shape — `Tray(backend::Handle)` over a `pub(super)` `Handle` — and the two paths differ only in which module the alias resolves to.

I swept both crates for platform-gated `pub` items rather than fixing only the six CI named: those six are the complete set, and no Windows-gated code declares any.

**CI is now green**, including all three `CLI Check` targets (Linux, macOS, Windows) and Clippy. Only `Test (Rust)` is still running.

Worth carrying into PR 3 and PR 4: `gglib-runtime` and `gglib-proxy` both have platform-gated code, so I'll check the gated regions by hand there rather than trusting a local `cargo fix` to have covered them.

PR #845, opened 2026-08-14
chore(proxy,axum,mcp,runtime): close the module trees in the four library crates

mmogr:
Third step of the dead-code sweep, one commit per crate. Same cascade as #844:
close the module tree so `dead_code` can see inside, then delete what it finds.

## What the compiler found

| Crate | Modules demoted | Dead code found |
|---|---|---|
| `gglib-mcp` | 8 of 8 | `MockFs::with_non_executable` |
| `gglib-axum` | 10 of 10 | `servers::start`, `servers::stop`, `sse::create_broadcaster` |
| `gglib-runtime` | 4 of 11 | none — all used, just over-public |
| `gglib-proxy` | 16 of 22 | `InitializeParams`, `ClientInfo`, `ToolsCallParams`, `INTERNAL_ERROR`, `SessionState::created_at` |

`servers::start`/`stop` are the path-based handlers for `/api/servers/{id}/start`
and `/stop`. Their own section comment calls them legacy; the mounted routes are
`start_body`/`stop_body`, and nothing routes these two.

The proxy findings answer a question this sweep had already queued for a later
PR. `InitializeParams` and `ToolsCallParams` were flagged as "investigate before
touching" because they looked like MCP wire types only tests parsed. They are:
`handle_initialize` never looks at `params`, and `tools/call` forwards
`request.params` as a raw `Value`. The structs and their two tests described a
shape nothing reads.

Closing the modules also exposed a dozen dead re-export lines that `pub` had been
hiding — `LockError`, `LockInfo`, `shutdown_signal`, `AgentRequestConfig`, and
nine names in `resolver/mod.rs` that everything inside `resolver/` reaches
through its defining module instead.

## Three things worth reviewing carefully

**1. Two lints want opposite things, and one has to go.** `gglib-mcp` inherits
`[workspace.lints]`, which includes clippy's `nursery`. `redundant_pub_crate`
fires on `pub(crate)` inside a private module and says use `pub`;
`unreachable_pub` fires on unreachable `pub` and says use `pub(crate)`. Demoting
a module is exactly what puts its contents in a private module, so both hold at
once and no visibility satisfies both.

I've allowed `redundant_pub_crate` at the workspace level, alongside the five
patterns already allowed there. It's the nursery one, and its advice is
counterproductive here: `pub` is precisely what stops `dead_code` looking, which
is the thing this arc exists to undo. `gglib-mcp`'s items therefore stay `pub`
and only its modules close — which loses nothing, because `dead_code` reasons
about effective visibility.

**This changes the final PR of the arc.** Enabling `unreachable_pub` workspace-wide
is only coherent with that allow in place.

**2. `unreachable_pub` has a false positive on re-export chains.** It flagged
`pub use lock::DaemonLock` and `pub use gglib_core::server_config::ServerConfigOptions`,
both of which `lib.rs` re-exports onward. Demoting either fails with E0365. Kept
`pub` with a note at each site. The compiler catching it is the point — nothing
here is applied on the lint's word alone.

**3. `JsonRpcRequest::jsonrpc` is kept deliberately**, with a targeted allow. It
is never read, but being a required field is what makes serde reject a request
that omits `jsonrpc` entirely, so deleting it would widen what the gateway
accepts. Separately worth a look: **nothing checks that value against `"2.0"`.**
The doc comment says it must be; no code enforces it.

## Method notes

`cargo fix` applies only a handful of `unreachable_pub` suggestions on crates
this size — 10 of 178 on `gglib-axum`. The suggestions are MachineApplicable with
exact byte spans; they just live in each diagnostic's `children` rather than on
the top-level span. I drove the edits from the JSON directly, asserting every
rewritten span is exactly `pub` and that the replacement is a visibility keyword.

Two traps that only `gglib-runtime` has, both caught before CI this time:

- **Features.** Its build pipeline is behind `cli`/`prebuilt`. A default-feature
  check reported `has_git`, `has_cmake`, `has_cpp_compiler` and `command_stdout`
  as dead and demoted them; with `cli` on, three are publicly re-exported and the
  build fails with E0364. Redone with `--all-features`, which is what CI runs.
- **Platforms.** 14 `pub` items sit in Linux/Windows-gated regions a macOS build
  never compiles — the blind spot that broke the tray modules in #844. Found by
  scanning for gated `pub` items rather than waiting for CI. Each takes the
  visibility of the twin this machine does compile.

## Verification

- `make pre-commit` passes
- `cargo clippy --all-targets --all-features -- -D warnings` clean across the workspace
- All four crates' test suites pass, doctests included
- `unreachable_pub = "deny"` added to `gglib-proxy`, `gglib-axum` and
  `gglib-runtime` (three of the eight crates that inherit nothing), each verified
  to fire by injecting a `pub fn` into a private module
- Six over-budget files gained 3–6 lines to rustfmt rewrapping longer
  signatures; recorded via the ratchet's `--update`


mmogr:
All three in 88529d09. Two corrections to flag, both verified rather than assumed.

### 1. Comments — you're right about the lint, and my diagnosis was worse than wrong

I tested it: with the root re-export present, `unreachable_pub` does **not** fire on the intermediate link. It tracks the chain correctly, exactly as you said.

What actually happened is narrower, and it's my tooling's fault rather than the lint's. Both sites were multi-name `use` statements where *some* names were unreachable:

```rust
pub use lock::{DaemonLock, LockError, LockInfo};
pub use gglib_core::server_config::{ServerConfigOptions, resolve_context_size};
```

`LockError`, `LockInfo` and `resolve_context_size` are named nowhere outside their crates — the lint was right about them. Its suggestion demotes the statement's single `pub` keyword, and my script applied that, taking the reachable name down with them. Confirmed by restoring the combined line and watching the warning point at column 58, which is `resolve_context_size`.

**One correction to the correction: it's E0365 at both sites, not E0364.**

```
crates/gglib-axum/src/lib.rs:49:18:   error[E0365]: `DaemonLock` is only public within the crate…
crates/gglib-runtime/src/lib.rs:53:25: error[E0365]: `ServerConfigOptions` is only public within the crate…
```

E0364 is the neighbouring case — re-exporting an item whose *definition* is `pub(crate)` — and it did appear earlier in this branch, on `get_num_cores`. I've used your wording with the code corrected, plus a sentence on the real mechanism, since "the lint is fine, my fix span was too wide" is the useful lesson for whoever automates this next.

### 2. `jsonrpc` validated — `allow` gone

Added where you suggested, before the notification branch so it covers those too. Two tests: one pins that a body omitting `jsonrpc` fails to deserialise (the half serde does), one that a well-formed `"1.0"` parses fine and needs the explicit check to be refused.

### 3. `test-support` gate — done, with a cost worth knowing

Mirrors `gglib-db`'s `test-utils`, plus a self dev-dependency to switch it on for `tests/`, which link the crate externally and so can't see `#[cfg(test)]`.

**It interacts with the `unreachable_pub = "deny"` this PR also adds.** Removing the public path in a default build leaves those items unreachable, so `cargo check -p gglib-proxy` with no features failed until I added a conditional allow on the module:

```
default features exit=101   # before
all features     exit=0     # which is why --all-features CI would not have caught it
```

Three items, not two — `LastLoadedSession` is a public field type of `StreamConfig`, so it loses reachability alongside. The allow is `cfg_attr`'d to apply only in the configuration where it's true, so the lint stays live for the `--all-features` build CI checks.

If you'd rather not pay that, dropping the gate is a two-line revert and I'm happy either way — flagging it because it's a genuine interaction between two things this PR does, not a detail.

### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` (default features) — clean, the state that was broken
- `make pre-commit` — passes
- Gate proven from the other side: `use gglib_proxy::StreamConfig` in `gglib-cli` gives E0432
- `handlers.rs` grew 526 → 564 for the check and its tests; recorded via the ratchet's `--update`

PR #846, opened 2026-08-14
chore(core): close the module tree in the crate fourteen others depend on

mmogr:
Fourth step of the dead-code sweep. `gglib-core` is on its own because it has
fourteen dependents and the widest blast radius in the workspace.

## Where the value turned out to be

Its **root** module tree really is its API — 20 of 21 root modules are named by
another crate, so there was nothing to win there. A level down is different: of
**133 `pub mod` declarations, 73 are named by nobody outside this crate.**

Whole subtrees move: every module under `ports/` except six, all but three under
`normalize/`, the seven inference and sampling modules under `request_pipeline/`,
and `domain::inference` — 1,563 lines of production code no consumer has ever
addressed by path.

Nothing became unreachable in the process, because of how this crate is built:
each `mod.rs` re-exports what it wants public, so `domain::InferenceConfig`
survives `domain::inference` going internal. That also explains a number worth
noting — only **22 items** needed demoting behind those 73 modules, against 174
in `gglib-axum`. The re-export layer was already doing the work; the module wall
was the missing part.

## `dead_code` found nothing

Like `gglib-runtime` in #845, this crate's contents are all used and only its
visibility was loose. Saying so plainly rather than implying the sweep finds
something everywhere it looks — three of the six crates audited so far have come
back clean, which is itself a result about this codebase.

## The one consumer change, and why the scan can't be trusted alone

`gglib-cli` imported:

```rust
use gglib_core::domain::agent::{AssistantContent, tool_types::ToolCall};
```

`ToolCall` is re-exported one level up, so the import now takes the re-export.
That line matters beyond the fix: the scan that picked these 73 modules reads
`gglib_core::a::b` paths, and **a module name nested inside braces is invisible
to it.** The compiler caught this one. It is the only reason it is not still
broken, and the reason the demotion was applied first and adjudicated second
rather than trusted up front.

## Three doctests, in their own commit

`utils::text` and `utils::timing` go internal, which breaks any doctest
importing through them — a doctest compiles as its own crate.
`format_duration_human` keeps an executable example because it is re-exported
from the crate root (`gglib-agent` already imports it that way) and only the path
was stale. The two `text` helpers have no consumer outside `gglib-core` at all,
so theirs become `text` blocks; the same cases are asserted for real in the
module's tests.

## A pre-existing finding, not from this branch

`cargo doc -D rustdoc::broken_intra_doc_links` **already fails on `main`**, with
18 unresolved links. This branch has 12 and introduces none — I diffed the two
sets to be sure. The count drops only because private modules are no longer
documented. Not fixed here; worth its own pass, since CI does not currently run
that denial.

Related: `gglib-core`'s published rustdoc will lose the 73 internal modules. That
is the intended outcome — they were never part of its API — but it is a visible
change if anyone reads those pages.

## Scope note

`unreachable_pub = "deny"` is *not* added here. `gglib-core` inherits
`[workspace.lints]`, and Cargo will not let a manifest hold both that and its own
lint keys — the same trade documented in #845. It is covered by the
workspace-wide lint at the end of this arc.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` (default features) — clean
- `make pre-commit` — passes
- `cargo test -p gglib-core --all-features` — passes; doctests 13 passing, up from 12
- No new broken intra-doc links, verified by diffing against `main`


mmogr:
Pushed 7088278c. Taking these in order of how right they were.

### The facade point is correct, and it was a real gap in my method

You're right that demoting modules while `lib.rs` and `domain/mod.rs` re-export the contents does nothing — the item stays reachable and `dead_code` stays quiet. Of the **316 names those two files re-export, 57 entries are named nowhere outside the crate.** Pruned.

Seven had to come back into `domain/mod.rs` — `RangeValues`, `capabilities_from_architecture`, `infer_from_chat_template`, `generation_config_candidates`, `parse_generation_config`, `declared_name`, `is_system_tag` — because they're reached internally as `crate::domain::X`. They're gone from `lib.rs`, which is where they were public API rather than convenience.

### What the pruning found: 8 more over-public items, and no dead code

You asked exactly what new dead code appeared. The answer is **none**, and I'd rather say that than manufacture a finding:

```
after pruning:  8 unreachable_pub  →  demoted
                0 dead_code  (--all-targets, as CI runs it)
                0 dead_code  (lib only, which reveals test-only items)
```

I spot-checked the pruned names to make sure that wasn't an artefact — `SEARCH_RESULTS_CAP` has 9 internal references, `ProfileNameError` 14, `parse_build_number` 8. They're live inside the crate; the re-exports were surplus API surface, not cover for anything unused.

So the win here is a smaller public API, not deletions. Real, but not what either of us expected.

### The lint — one correction to the premise

I didn't claim to have added it. The commit says the opposite: *"`unreachable_pub` is not denied for this crate: it inherits `[workspace.lints]`… covered by the workspace-wide lint at the end of this arc."* No `Cargo.toml` change was in the diff, consistent with that.

Pulling it forward is a good call though, and it's done: `unreachable_pub = "deny"` in `[workspace.lints.rust]`, with `#![allow(unreachable_pub)]` in the six crates you named.

Three more needed attention that the list missed — `gglib-agent`, `gglib-bootstrap` and `gglib-integration-tests` fail in their **`tests/` directories**, which are separate crates a `lib.rs` attribute can't reach. I cleaned those fixtures (27 items) rather than adding suppressions.

Verified both directions: a `pub fn` injected into a private module of `gglib-core` fails the build; the same injection into `gglib-hf` doesn't.

### Doc links — you're right, and I'd introduced one

My "12 vs 18" was measured **without** `--document-private-items`, which is exactly the artefact you describe. In CI's configuration it's **63 either side** — unchanged, as you said.

Re-measuring properly also caught something I'd missed: pruning `validate_name` from the `domain` facade broke `[crate::domain::validate_name]` in `settings.rs`. Fixed to point at `domain::inference_profile::validate_name`. The unresolved-link set is now byte-identical to `main`.

Those 63 all pre-date this branch and CI doesn't currently deny them — still worth its own pass.

### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo doc --document-private-items` — 63 unresolved, byte-identical to main
- `make pre-commit` — passes

PR #847, opened 2026-08-14
chore(workspace): delete the code nothing reads

mmogr:
Fifth step of the dead-code sweep, and the first that is purely deletion. **534
lines removed, nothing added but comments.**

## The dead duplicate — 250 lines

`gglib-runtime/src/proxy/models.rs` held the OpenAI request and response types
— `ChatCompletionRequest`, `ChatMessage`, `ToolCall`, `ErrorResponse` — in a
near-identical copy of the live set in `gglib-proxy/src/models.rs`. Nothing in
the workspace imported it; the three places in that crate wanting `ErrorResponse`
reach for `gglib_proxy::models::ErrorResponse`. Both copies have existed since
the first commit. This one was touched once, in #439, and never again.

Worth stating why it survived everything:

| Check | Why it missed |
|---|---|
| `dead_code` | `pub mod proxy` → `pub mod models` made every item externally reachable, so the compiler had to assume another crate was the caller |
| `unreachable_pub` | same reason — the items *were* reachable |
| text search | every name in it also exists in the live copy, so the symbols look referenced wherever you grep |

Reachable, duplicated, and unreferenced at once. That combination is exactly what
a module-by-module visibility pass exposes and nothing else does.

## The 27 items — 284 lines

Items no code, test, doc or script names anywhere; the definition is the only
mention. Unused builders lead (`with_static_dir`, `with_host`,
`with_allowed_origins`, `with_extra_args`, `with_send_timeout`), then the
database factory's `create_pool` / `create_test_pool` / `new_in_memory_blocking`,
`cleanup_zombie_runs` (callers bypass the wrapper for
`gglib_db::cleanup_zombie_benchmark_runs`), and sixteen others.

Each was verified individually rather than trusted from the scan — the scanner
has been wrong before in this arc, and a false positive here deletes live code.

**The cascade needed two rounds.** Removing the first set stranded three more,
which the compiler then found: `LogBuffer::new` (reached only via `init_server`;
the derived `Default` is what the surviving path uses), `str_eq` (a const helper
that existed solely to compute `GIT_DIRTY`), and a `MessageRole` import.

## Docs and build script

- `gglib-build-info`'s README documented `GIT_DIRTY` in an ASCII diagram, a
  constants table and a fallback list — and its `build.rs` still emitted the
  `VERGEN_GIT_DIRTY` env var that constant was the only reader of. Both cleaned.
- `gglib-sse`'s README and an **intra-doc link** in `gglib-proxy`'s pointed at
  `subscribe_with_hydration`; both now name the `_until` variant production calls.

The unresolved intra-doc-link set is byte-identical to `main` — checked in CI's
`--document-private-items` configuration this time.

## One item deliberately left

`DOWNLOAD_SYSTEM_ERROR` is unreferenced in Rust, but
`src/hooks/useDownloadSystemStatus.ts` still listens for the event string it
holds. Deleting the constant alone would leave the frontend waiting on an event
whose name no longer exists in the codebase. Both ends belong in one change —
that's the event-lane PR, not this one.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` (default features) — clean
- `cargo test --workspace --all-features` — 80 suites pass
- `cargo doc --document-private-items` — unresolved-link set identical to `main`
- `make pre-commit` — passes


mmogr:
All four in 9ff79670. The vergen catch was a good one — I'd verified the wrong half of that change.

### 1. `.dirty(false)` — confirmed, and worse than "left behind"

I checked the vendored source rather than take it on trust:

```rust
// vergen-gix-9.1.0/src/gix/mod.rs:258
/// Emit the dirty state of the git repository
/// ```text
/// cargo:rustc-env=VERGEN_GIT_DIRTY=(true|false)
/// ```
/// Optionally, include untracked files when determining the dirty status…
pub fn dirty(&mut self, include_untracked: bool) -> &mut Self
```

So the bool only decides whether untracked files count towards dirtiness — the call itself is what turns emission on. My previous commit removed the *fallback* emission while the main path carried on emitting for a constant that no longer existed. Removed from the chain, and the comment now records which of the two things that argument does, so the next person doesn't repeat it.

### 2. `FORMAT_THINK_TAG` — restored, and my process was at fault

ADR 0001 line 78: *"`format:think-tag` is reserved in `normalize::tags` but has no parser yet."* The constant's **own doc comment** says it too, in the sentence right above the line I deleted:

```rust
/// V1 reserves this constant for forward compatibility; the corresponding
/// parser is delivered in a follow-up PR.
pub const FORMAT_THINK_TAG: &str = "format:think-tag";
```

A reference count of zero isn't sufficient grounds when the item's whole purpose is to be reserved. I verified each of the 27 items for *references* and didn't read what several of them said about themselves. That's the lesson I'm taking rather than just the fix.

### 3. Empty `impl` blocks — removed

`impl ServerSummary {}` and `impl HfModelSummary {}`, left when `parsed_model_id` and `repo_ref` turned out to be their only members.

### 4. Event lane — agreed, with one correction to the diagnosis

You're right that the frontend doesn't read the Rust constant; `useDownloadSystemStatus.ts` has its own `'download-system:error'` literal. So deleting the constant would have left the listener in place and the UI stuck on `initializing` regardless — it would have removed the *record* of the problem, not the problem.

One nuance on "all of `gglib-tauri`'s events module is orphaned" — I audited all 16 constants end to end:

| | emitted in Rust | listened for in TS |
|---|---|---|
| `DOWNLOAD_SYSTEM_ERROR` | **0** | 1 |
| the other 15 | 1–4 each | 1 each |

So the module isn't orphaned; it has exactly one gap. The structural point behind your claim holds though: `pub mod events;` in that crate's `lib.rs` is what stops `dead_code` from saying so, and **`gglib-tauri` is still unswept** — it has neither the workspace lints nor its own `unreachable_pub`, so it's the one crate this arc hasn't accounted for.

Plan updated: no event-lane deletion PR; the tracking issue picks it up after merge.

### Verification

- `cargo clippy` — clean in both feature modes
- `cargo build -p gglib-build-info` — compiles, version constants intact
- `cargo doc --document-private-items` — 63 unresolved links, byte-identical to `main`
- `make pre-commit` — passes

PR #885, opened 2026-08-20
Inference profile selection for gglib chat, q and serve

mmogr:
Inference profiles were selectable only by an HTTP client, via a `{model}:{profile}` suffix on the requested model id. `gglib chat`, `gglib q` and `gglib serve` could not select one at all — the profile-aware primitive existed in core, but nothing threaded a profile to the CLI's resolve call, and the router that reads the suffix lived inside `gglib-proxy` where only the proxy could reach it.

This adds both selection forms to all three commands, and removes the duplication that caused the gap rather than adding to it.

**It also fixes a live bug found during review**, which is the largest behaviour change here — see the `fix(...)` commit below and read it before the feature commits.

## The refactor block (commits 1–7)

Four preparatory commits make room: five of the files this work touches sit on the 300 LOC ratchet, where the gate fails on any growth at all, and `commands.rs` alone needed six lines it did not have. Then three commits move shared code: the `{model}:{profile}` router into `gglib-core`, the two duplicate `explain` implementations onto one kernel, and field provenance out of the CLI and pinned-launch resolutions.

Every one of these is provably no-behaviour-change. The router move is verifiable in the strongest form available — `integration_inference_profiles.rs` is untouched and its 13 tests still pass, and they assert on the body the upstream actually receives.

## `gglib serve`'s sampling flags were inert

`gglib serve 7 --temperature 0.9` resolved the value, printed "Temperature: 0.9" in its banner, and sampled at whatever the model and global layers said. The flags rode the pinned model's launch options into `ServerConfig::inference_config`, whose own doc says it is "written by `build_server_config` and read by nobody" and warns the next reader not to wire anything to it. gglib emits no sampler flags to llama-server at all (ADR 0003/0004), so those options were never a path to the sampler.

This restores what `docs/sampling.md` already promised. Three consequences are handled rather than left to be discovered:

- **Re-invocation** against a running proxy was idempotent success comparing only the pinned model's *name*, so the override would have been printed and discarded. It now refuses and names `gglib proxy stop`.
- **Client budgets.** `--max-tokens` and `--reasoning-budget-tokens` are the two values the trust gate deliberately lets an untrusted client keep. The operator rung sits above them, so `serve` now says at launch that client budgets on this endpoint are capped. Not a refusal — `gglib proxy` has always behaved this way, and excluding budgets here would make the two commands differ.
- **The banner** printed a five-rung stored resolution as though it were the six-rung one requests actually get. It now prints only what the operator passed.

## The feature (commits 9–13)

`--profile coding` and `qwen3.6:coding` both work on `chat`, `q` and `serve`; passing both is an error rather than a precedence rule. On `serve` the profile is an endpoint **default**, carried to the daemon by name so editing it later takes effect without a restart, and a client can still override per request.

A separate commit warns when the temperature-coupled trio discards a flag — `--profile chat --presence-penalty 1.2` silently dropped the penalty. That footgun predates profiles, so it lands before them and is verifiable on its own.

## Notes

- `gglib proxy` deliberately gets no `--profile`: it serves every model, so a single default has no model in scope.
- `serve`'s positional widens from `u32` to a string. Backward compatible, and it finally lets `serve` name a model like every other command.
- Known gap, filed rather than fixed: `ConversationSettings` restores temperature/top_p on `chat --continue` but has no profile field, so a resumed session loses its profile.


mmogr:
Post-review update: two blockers found by adversarial review of the diff, both fixed in the two commits just pushed.

**`gglib proxy` regression.** The re-invocation guard keyed on "an override is present" rather than comparing it, and sat in the shared start path — so re-attaching after a detach, and (with `proxy_autostart`) the *first* `gglib proxy --temperature …` of a session, both started returning 409. Now compares against what the run was actually started with. This also closed a hole the original missed: `serve --profile` against a running endpoint returned success and applied nothing.

**`gglib chat --continue N --profile X` was unusable** — the invocation this PR's own docs advertise. The conversation record stored the identifier with its `:profile` suffix, so every resume replayed a suffix the user never typed and collided with their flag; deleting the profile made the conversation permanently unresumable. Selection now runs before persistence, and a resume treats a stored suffix as overridable.

Also: the banner printed 7 of 18 settable fields; a per-request WARN became once-per-run; a flag beside a mistyped suffix reported a missing *model* instead of the bad *profile*; `gglib q --model qwen:coding` sent the suffixed name upstream; and a docs sentence claimed `serve` warns on discarded coupled flags when it cannot honestly do so.

Known gap, not fixed: `discarded_from_rung` has a false negative for non-finite values — `--presence-penalty inf` serialises to null, drops out of the JSON patch, and is never reported as discarded.

mmogr:
Both remaining known gaps are now fixed, and the live `serve` verification is done.

**Live verification** (model 1, 27B, daemon restarted from this branch — the running one was still on `main`, which would have exercised none of the daemon-side changes):

| | |
|---|---|
| Budget warning | printed at launch |
| **Operator sampling reaches llama-server** | client sent `max_tokens: 4096`, got **512**, `finish_reason: length` |
| Re-invocation, different sampling | 409, names `gglib proxy stop` |
| Re-invocation, identical sampling | re-attaches, no 409 |

The middle row is the one that matters: before the `fix(...)` commit that request returned up to 4096 tokens, because the flag was inert. Budgets are the one class of client value that survives the trust gate, so beating one proves the override path end to end.

**Two fixes since:**

- `fix(cli): print sampling values the way the operator typed them` — the patch-based banner printed `--temperature 0.1` as `0.10000000149011612` (f32 widened through JSON). Only the live run surfaced it; nothing asserts on banner text.
- `fix(core): report a sampling flag that is named but cannot be sent` — closes the `discarded_from_rung` false negative flagged above. A non-finite value is dropped by JSON serialisation, so it never appeared in the caller's patch and was skipped before the coupling rule ran. Now reported unconditionally, including when the caller's rung won the ladder — winning is not the same as being sent. The float list is explicit with two drift guards, one of which I verified fails when a field is removed from it.

